PR #41, opened 2023-12-18
Fix misc.R bugs

Travyse:
Fixed errors that prevented `load_regional_univariate_data` from working. I tested this locally and the output matches the output that the script produced before I refactored anything.

### DIFF
--- a/R/misc.R
+++ b/R/misc.R
@@ -80,8 +80,6 @@ prepare_data_list <- function(geno_bed, phenotype, covariate, region, imiss_cuto
         mutate(
           covar = map(covariate_path, ~read_delim(.x,"\t")%>%select(-1)%>%na.omit%>%t()),
           Y = map2(phenotype_path,covar, ~{
-            print(.x)
-            print(region)
             y_data <- tabix_region(.x, region)%>%select(-4)%>%select(rownames(.y))%>%t()%>%as.matrix
             return(y_data)
           }),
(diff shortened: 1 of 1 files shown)


gaow:
Thanks @Travyse not too bad -- i was so close to fixing up all of it! It's nice that you are able to test and verify. I think it would make sense if you could use chatgpt to generate some unit test for this. 

For now, could you please clean up your code removing various `print` statement that you had there for debug?

Travyse:
@gaow Ah, I forgot about all of those. I pushed some changes without the debugging print statements. I will try to use ChatGPT/myself for some unit testing asap

gaow:
@Travyse merged for now! Thanks we'll make a container for this tomorrow after I finalize and test with @AnqiWang2021 's new PR, and resume running production pipeline. 

Moving forward we should :

1. Start adding unit tests for this and various other functions. 
2. As discussed on slack, add a new parameter `keep_samples` to only load a list of specified samples, for CV purpose. 

@Travyse do you propose any other codes to be refactored and improved? You may noticed we added some wrappers to popular tools `susie_wrapper` `mash_wrapper` and `mrmash_wrapper` make these tools much easier to use. You don't have to worry about those at least not for now. Once we are done with unit testing and refactor we can look into actual TWAS-MR analysis with CV on the cloud, to complete that result section of the paper.

Travyse:
@gaow At the moment, I don't have any propositions for refactoring currently (that may change as I complete more unit tests). I am using ChatGPT as mentioned to help generate the data for some of the unit tests. Should I track the prompts used (for citation purposes or something)?

gaow:
Thank you @Travyse it's okay I don't think we need to track the prompts. However it's important you make sure every single line of code it writes for you makes 100% sense to you. It makes subtle mistakes all the time is my experience and sometimes it repeats mistakes it previously made. You need to be super picky and grumpy about codes it gives you. Of course comprehensive unit testing will help a lot too

PR #43, opened 2023-12-18
Add keep_samples parameter in misc.R

Travyse:
Added a `keep_samples` parameter to the `load_genotype_data` function in misc.R


gaow:
Thanks @Travyse a couple of comments:

1. we need to have the keep_samples paramter to the interface, ie, the exported function like load_univariate_data, load_multivariate_data or load_finemapping_data I cannot remember the exact function names but you get the idea. 
2. for keep_samples are you relying on filtering genotypes, then phenotype and covariates will be dropped correspondingly? I'm reading the code. i can see  X matrix are removed of samples not existing in Y. But I think we assume (or our pipeline assumes) that Y and covar data have the same samples.  My point is that we should make sure that X, Y and covar all match with each other including the ordering of samples
3. https://github.com/cumc/pecotmr/blob/main/R/misc.R#L124 this line and the next line are duplicated.

gaow:
I'll merge it for now and try to edit it a bit

gaow:
These issues above are addressed in current `main`. Please test both via your own test/debug recipe and via unit testing.

PR #44, opened 2023-12-19
Implementation on the grid computation

Chunmingl:


R/mrmash_wrapper.R
```diff
@@ -181,35 +200,11 @@ mrmash_wrapper <- function(X,
 #'
 ### Function to compute initial estimates of the coefficients from group-lasso
 compute_coefficients_glasso <- function(X, Y, standardize, nthreads, Xnew=NULL, version=c("Rcpp", "R")) {
```
gaow, line 202: since version parameter is no longer used should we remove it?

R/mrmash_wrapper.R
```diff
@@ -115,15 +117,32 @@ mrmash_wrapper <- function(X,
...
 
   if (is.null(prior_data_driven_matrices) && !isTRUE(prior_canonical_matrices)) {
     stop("Please provide prior_data_driven_matrices or set prior_canonical_matrices=TRUE.")
   }  
+  
+    
+  Y_has_missing <- any(is.na(Y))
+    
+  if(Y_has_missing && B_init_method=="glasso"){
+        stop("B_init_method=glasso can only be used without missing values in Y")
```
gaow, line 129: Should we provide some more hints and tips to users to guide them what to do in this situation?

PR #47, opened 2023-12-25
ctwas_wrapper draft and example

Chunmingl:


gaow:
Merging as is for now until we talk to the other developers as discussed on slack

PR #48, opened 2023-12-25
Add unit tests for misc.R

Travyse:
- Added unit tests for almost every function in `misc.R`
  - Still need to unit test `load_regional_multivariate_data`
- Made minor changes to `misc.R` as necessary
- Made minor change to `LD.R` to allow for a clearer way to check for `bim_path`


R/LD.R
```diff
@@ -90,7 +90,8 @@ bt_intersect <- function(df, region_strings) {
     # Get file paths between row_start and row_end
     file_paths <- df$path[which(df$chrom == rows$start_row$chrom & df$start >= rows$start_row$start & df$start <= rows$end_row$start)]
     # check if there exits bim_path, if the path is present, we use directly, else is NULL
-    if(!is.null(df$bim_path))
+    #if(!is.null(df$bim_path))
```
gaow, line 93: what you did seems correct so please remove this line for good.

R/misc.R
```diff
@@ -180,14 +186,13 @@ add_Y_residuals <- function(data_list, conditions, y_as_matrix = FALSE, scale_re
                      unnest(Y_resid) %>%
                      as.matrix()
     colnames(Y_resid_matrix) <- conditions
-    data_list$Y_resid <- Y_resid_matrix
+    data_list$Y_resid <- list(Y_resid_matrix)
```
gaow, line 189: Please revert this line for now. I believe what we previously had is correct but i'm yet to test that --- it is going to be relevant to the mvSuSiE pipeline which exists but have not yet been tested to work with pecotmr. You can revert it for now. My team will take care of testing it.

R/misc.R
```diff
@@ -302,104 +309,4 @@ load_regional_multivariate_data <- function(matrix_y_min_complete = NULL, # when
         chrom = dat$chrom,
         grange = dat$grange
         ))
-}
```
gaow, line 305: @Travyse this is some code I wrote today but you overwrite and deleted it. Please recover it. Before you send in PR you should sycn your branch first. Otherwise you'll overwrite other peoples' updates.

PR #49, opened 2023-12-29
mrmash_wrapper with cv, grid on the fly, vignettes update

Chunmingl:


R/mrmash_wrapper.R
```diff
@@ -301,6 +325,30 @@ compute_w0 <- function(Bhat, ncomps) {
...
+#         names(datadriven_mats_filt) <- names(datadriven_mats)
+            
+#     } else {
+#           # if no cross validation 
+#         datadriven_mats_filt <- lapply(datadriven_mats, function(x, to_keep){x[to_keep, to_keep]}, tissues_to_keep)
+#     }
+#     return(datadriven_mats_filt)
+    
+# }
+
```
gaow, line 351: @Chunmingl are these commented out codes still useful? If you are not sure about it please discuss with us on slack. if you are sure they are not useful please remove them.

R/mrmash_wrapper.R
```diff
@@ -356,4 +404,41 @@ autoselect_mixsd <- function(gmin, gmax, mult=2){
...
+        
+#     }
+    
+#     return(res_all_folds)
+    
+# }
+
+
+#' @importFrom mr.mash.alpha compute_univariate_sumstats
+sumstat <- function(X, Y, standardize, nthreads){
```
gaow, line 439: this wrapper function seems not useful and the function name is also confusing (it is not a verb). Please remove this wrapper and simply use `compute_univariate_sumstats` unless you have a good reason to not do that.

R/twas.R
```diff
@@ -92,19 +92,33 @@ twas_weights_cv <- function(X, Y, fold = NULL, sample_partitions = NULL, weight_
...
+            folds <- cut(seq(1, nrow(X)), breaks = fold, labels = FALSE)
+            sample_partition <- data.frame(Sample = sample_names[sample_indices], Fold = folds, stringsAsFactors = FALSE)
+        }
     } else if (!is.null(sample_partitions)) {
         if (!all(sample_partitions$Sample %in% sample_names)) {
             stop("Some samples in 'sample_partitions' do not match the samples in 'X' and 'Y'.")
         }
         folds <- sample_partitions$Fold
         sample_partition <- sample_partitions
+        fold <- max(unique(sample_partition$Fold))
```
gaow, line 116: It's safer to use `length(unique(sample_partition$Fold))`?

R/twas.R
```diff
@@ -130,13 +152,14 @@ twas_weights_cv <- function(X, Y, fold = NULL, sample_partitions = NULL, weight_
                 args <- weight_methods[[method]]
                 if (method %in% multivariate_weight_methods) {
                     # Apply multivariate method to entire Y for this fold
-                    weights_matrix <- do.call(method, c(list(X = X_train, Y = Y_train), args))
+                    weights_matrix <- do.call(method, c(list(X = X_train, Y = Y_train, sample_partition=sample_partition), args))
```
gaow, line 155: why do you still need sample_partition in the method itself?

R/mrmash_wrapper.R
gaow: @Chunmingl here is a comment you havent yet addressed

R/twas.R
```diff
@@ -65,17 +65,20 @@ twas_z <- function(weights, z, R=NULL, X=NULL) {
 #' @importFrom foreach %dopar%
 #' @importFrom doParallel registerDoParallel
 #' @export
-twas_weights_cv <- function(X, Y, fold = NULL, sample_partitions = NULL, weight_methods = NULL, seed = NULL, num_threads = 1) {
+twas_weights_cv <- function(X, Y, fold = NULL, sample_partitions = NULL, prior_data_driven_matrices=NULL, weight_methods = NULL, seed = NULL, num_threads = 1) {
```
gaow, line 68: I wonder if you can try to implement this interface:

`twas_weights_cv(X, Y, fold = NULL, sample_partitions = NULL, weight_methods = NULL, seed = NULL, num_threads = 1, ...)`


where `...` is left for other parameters specific to methods. Then you can match `...` with parameters such as `prior_data_driven_matrices`. for example, https://stackoverflow.com/questions/5890576/usage-of-three-dots-or-dot-dot-dot-in-functions

gaow: Similarly this interface can also be added to `twas_weights` function.

R/mrmash_wrapper.R
Chunmingl: It has been cleared and updated

R/twas.R
Chunmingl: ok, it is now fixed

R/mrmash_wrapper.R
Chunmingl: it's now updated

R/twas.R
Chunmingl: It's now been removed

PR #53, opened 2024-01-04
Use snpStats read.plink for load_genotype_data

Travyse:
I reimplemented `load_genotype_data` as `load_genotype_data_snpStats`. It uses `read.plink` as discussed. I've also added unit tests to compare outputs of the two functions. As discussed with @gaow, I'm using some of the sample data he's provided me but they're maybe too large for unit tests. I will shorten the data after this push.


Travyse:
Also, I still need to figure out how to load the necessary libraries so that some of the tests don't fail. Aside from that, there's also the `load_regional_association_data` tests that fail (which I need to discuss with @gaow).

PR #60, opened 2024-01-16
fsusie update

hsun3163:
add
1. A function to save the phenotype info before sending the pheno object into processing
2. A wrapper for fsusie for future interface changes so that the core function was not exported and allows future updates.


gaow:
I change function and variable names https://github.com/cumc/pecotmr/commit/d3960765fe3874e5be5e7207508bd8f46a2b2165 will do the same in the xqtl-pipeline repo

PR #64, opened 2024-01-22
Add unit tests

Travyse:
Made minor changes to scripts (compute_qtl_enrichment) and added unit tests. Added placeholders for `file_utils`.

### DIFF
--- a/R/raiss.R
+++ b/R/raiss.R
@@ -24,7 +24,7 @@
 #' # result <- raiss(ref_panel, known_zscores, LD_matrix, lamb = 0.01, rcond = 0.01, R2_threshold = 0.6, minimum_ld = 5)
 raiss <- function(ref_panel, known_zscores, LD_matrix, lamb = 0.01, rcond = 0.01, R2_threshold = 0.6, minimum_ld = 5) {
   # Check that ref_panel and known_zscores are both increasing in terms of pos
-  if (is.unsorted(known_zscore$pos)|| is.unsorted(known_zscore$pos)) {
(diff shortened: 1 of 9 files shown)


R/raiss.R
```diff
@@ -223,9 +224,9 @@ invert_mat_eigen <- function(mat, tol = 1e-3){
     
     eigen_mat <- eigen(mat)
     L <- which(cumsum(eigen_mat$values) / sum(eigen_mat$values) > 1-tol)[1]
-    mat_inv <- eigen_mat$vectors[,1:L] %*% 
+    mat_inv <- if(!is.na(L)) eigen_mat$vectors[,1:L] %*%
```
gaow, line 227: @Travyse if L is NA we should throw and error? we should not return a matrix that is not processed.

R/raiss.R
```diff
@@ -24,7 +24,7 @@
 #' # result <- raiss(ref_panel, known_zscores, LD_matrix, lamb = 0.01, rcond = 0.01, R2_threshold = 0.6, minimum_ld = 5)
 raiss <- function(ref_panel, known_zscores, LD_matrix, lamb = 0.01, rcond = 0.01, R2_threshold = 0.6, minimum_ld = 5) {
   # Check that ref_panel and known_zscores are both increasing in terms of pos
-  if (is.unsorted(known_zscore$pos)|| is.unsorted(known_zscore$pos)) {
+  if (is.unsorted(known_zscores$pos)|| is.unsorted(known_zscores$pos)) {
```
gaow, line 27: looks like Haochen made a mistake when he modified the code. One of it should be `ref_panel`

PR #66, opened 2024-01-23
update enrichment and coloc

rfeng2023:


gaow:
Thanks @rfeng2023 I will write a more detailed review on Thursday so you can make some changes

R/encoloc.R
```diff
@@ -31,9 +31,9 @@ xqtl_enrichment_wrapper <- function(gwas_finemapped_data, xqtl_finemapped_data,
     # Process GWAS data
     gwas_pip <- list()
     for (file in gwas_finemapped_data) {
-        gwas_data <- readRDS(file)[[gwas_finemapping_obj]]
+        gwas_data <- if (!is.null(gwas_finemapping_obj)) readRDS(file)[[gwas_finemapping_obj]] else readRDS(file)
         pip <- gwas_data$pip
-        names(pip) <- gwas_data$variant_names
+        #names(pip) <- gwas_data$variant_names #FIXME: need to check the new results from gwas susie_rss
```
gaow, line 36: what you can do is `if (!is.null(gwas_data$variant_names)) names(pip) <- gwas_data$variant_names `?

R/encoloc.R
```diff
@@ -93,12 +88,14 @@ xqtl_enrichment_wrapper <- function(gwas_finemapped_data, xqtl_finemapped_data,
...
+                          enrich_res ) {
     # Load and process GWAS data
-    gwas_lbf_matrices <- lapply(gwas_files, function(file_path) {
-        gwas_data <- readRDS(file_path)[[gwas_finemapping_obj]]
+    gwas_lbf_matrices <- lapply(gwas_files, function(file) {
+        gwas_data <- if (!is.null(gwas_finemapping_obj)) readRDS(file)[[gwas_finemapping_obj]] else readRDS(file)
         gwas_lbf_matrix <- as.data.frame(gwas_data$lbf_variable)
-        colnames(gwas_lbf_matrix) <- gwas_data$variant_names
+        gwas_lbf_matrix <- gwas_lbf_matrix[gwas_data$sets$cs_index,]
+        #colnames(gwas_lbf_matrix) <- gwas_data$variant_names #FIXME: need to check the new results from gwas susie_rss
```
gaow, line 98: same as comment above.

R/encoloc.R
```diff
@@ -113,20 +110,34 @@ coloc_wrapper <- function(xqtl_file, gwas_files,
                                 mutate(across(everything(), ~replace_na(., 0)))
 
     # Process xQTL data
-    xqtl_data <- readRDS(xqtl_file)[[xqtl_finemapping_obj]]
-    xqtl_lbf_matrix <- as.data.frame(xqtl_data$lbf_variable)
+    xqtl_data <- if (!is.null(xqtl_finemapping_obj)) readRDS(xqtl_file)[[1]][[xqtl_finemapping_obj]] else readRDS(xqtl_file)[[1]]
+    xqtl_lbf_matrix <- as.data.frame(xqtl_data$susie_result_trimmed$lbf_variable)
+    xqtl_lbf_matrix <- xqtl_lbf_matrix[xqtl_data$susie_result_trimmed$sets$cs_index,]
```
gaow, line 115: We should not limit this to variants inside of CS. We need to use all variants.

R/encoloc.R
```diff
@@ -113,20 +110,34 @@ coloc_wrapper <- function(xqtl_file, gwas_files,
...
+    combined_gwas_lbf_matrix <- combined_gwas_lbf_matrix[, common_colnames, drop = FALSE] %>% as.matrix
 
     # Report the number of dropped columns from xQTL matrix
     num_dropped_cols <- length(setdiff(colnames(xqtl_lbf_matrix), common_colnames))
     message("Number of columns dropped from xQTL matrix: ", num_dropped_cols)
 
     # Return the processed data for now
-    # FIXME: To complete this, we need to apply colocalization analysis on these single effect logBF and the enrichment analysis results computed by `xqtl_enrichment_wrapper`
-    return(list(xqtl_lbf_matrix = xqtl_lbf_matrix, combined_gwas_lbf_matrix = combined_gwas_lbf_matrix))
+    # FIXME: do we only keep trimmed lbf_matrix or not
```
gaow, line 139: we don't keey any lbf_matrix. I don't see where we would need it for any of the downsteeams analysis.

R/encoloc.R
```diff
@@ -113,20 +110,34 @@ coloc_wrapper <- function(xqtl_file, gwas_files,
...
     # Report the number of dropped columns from xQTL matrix
     num_dropped_cols <- length(setdiff(colnames(xqtl_lbf_matrix), common_colnames))
     message("Number of columns dropped from xQTL matrix: ", num_dropped_cols)
 
     # Return the processed data for now
-    # FIXME: To complete this, we need to apply colocalization analysis on these single effect logBF and the enrichment analysis results computed by `xqtl_enrichment_wrapper`
-    return(list(xqtl_lbf_matrix = xqtl_lbf_matrix, combined_gwas_lbf_matrix = combined_gwas_lbf_matrix))
+    # FIXME: do we only keep trimmed lbf_matrix or not 
+    # COLOC function 
+    coloc_res <- coloc::coloc.bf_bf(xqtl_lbf_matrix, combined_gwas_lbf_matrix, p1 = enrich_res$`Alternative (coloc) p1`, p2 = enrich_res$`Alternative (coloc) p2`, p12 = enrich_res$`Alternative (coloc) p12`)
```
gaow, line 141: To write a proper R package please use 

```
#' @importFrom coloc coloc.bf_bf
```
 in the documentation string, then, add 

```
coloc
``` 

to `DESCRIPTION` file for the dependent packages, and finally run

```
devtools::document()
```

to update the documentation.

R/encoloc.R
```diff
@@ -113,20 +110,34 @@ coloc_wrapper <- function(xqtl_file, gwas_files,
...
+    return(list(xqtl_lbf_matrix_trim = xqtl_lbf_matrix, combined_gwas_lbf_matrix_trim = combined_gwas_lbf_matrix, coloc = coloc_res, enrichment = enrich_res))
```
gaow, line 142: also you don't have to return enrichment result again. let's just return coloc_res

R/encoloc.R
```diff
@@ -93,12 +88,14 @@ xqtl_enrichment_wrapper <- function(gwas_finemapped_data, xqtl_finemapped_data,
 #' @export
 coloc_wrapper <- function(xqtl_file, gwas_files, 
                           gwas_finemapping_obj = "susie_fit", 
-                          xqtl_finemapping_obj = "susie_fit") {
+                          xqtl_finemapping_obj = "susie_fit",
+                          enrich_res ) {
```
gaow, line 92: Let's use p1, p2 and p12 instead of enrich_res as the input parameter here.

gaow: and you can set default the same as `coloc` default.

R/encoloc.R
gaow: For the time being, can we limit it to those rows in this matrix where `$V>0`?

R/encoloc.R
```diff
@@ -109,13 +110,13 @@ coloc_wrapper <- function(xqtl_file, gwas_files,
     combined_gwas_lbf_matrix <- bind_rows(gwas_lbf_matrices) %>%
                                 mutate(across(everything(), ~replace_na(., 0)))
 
+
     # Process xQTL data
     xqtl_data <- if (!is.null(xqtl_finemapping_obj)) readRDS(xqtl_file)[[1]][[xqtl_finemapping_obj]] else readRDS(xqtl_file)[[1]]
     xqtl_lbf_matrix <- as.data.frame(xqtl_data$susie_result_trimmed$lbf_variable)
```
gaow, line 116: @rfeng2023 in this case you will set `xqtl_finemapping_obj = "susie_result_trimmed"` and not use this, Just `xqtl_data$lbf_variable` is good enough? Remember in a function for an R package we never want hard coded stuff like this.

PR #69, opened 2024-01-25
Create the wrapper for fsusie

hsun3163:
Now the wrapper do 3 things after running fsusie
1. filter out dummy cs
2. Create the sets list for fsusie
3. calculate the correlation between cs for the fsusie

All three functions has been tested, but need to be tested in the pecotmr container along with the newest version of fsusie


gaow:
@Travyse no need to unit test on this file for now because eventually we will work with `fsusieR` packing and move some of these into that package.

PR #73, opened 2024-01-26
fsusie with susie_post_process

hsun3163:


hsun3163:
Change the alpha to work with susie_post_process

PR #87, opened 2024-01-30
fix LD.R and allele_qc.R

AnqiWang2021:


R/LD.R
```diff
@@ -182,8 +182,24 @@ extract_LD_for_region <- function(LD_matrix, variants, region, extract_coordinat
 
 # Create a combined LD matrix from multiple matrices
 create_combined_LD_matrix <- function(LD_matrices, variants) {
-    # Create a block matrix with correct names
-    combined_LD_matrix <- as.matrix(do.call(bdiag, LD_matrices))
+    # Extract unique variant names from the list of variants
+    unique_variants <- unique(variants$variants)
```
gaow, line 186: I worry that the unique() function will completely mess up the order of variants. I think the issue you are trying to address here is very specific. We should not use this. We should just address the one variant that joins the two matrices.

R/LD.R
```diff
@@ -241,6 +256,7 @@ load_LD_matrix <- function(LD_meta_file_path, region, extract_coordinates = NULL
 
     # Create a combined LD matrix
     combined_LD_matrix <- create_combined_LD_matrix(combined_LD_matrices, combined_LD_variants)
+    combined_LD_variants <- combined_LD_variants[match(rownames(combined_LD_matrix),combined_LD_variants$variants),]
```
gaow, line 259: Here my undersatnding is that the difference is just one variant? then we should not match it like this. We should address it specifically and clearly.

gaow:
@AnqiWang2021 these lines in `process_LD_matrix` can be removed, because I have dealt with this previously when we load the meta data:

```
    # Check whether the LD_file_path and bim_file_path exist or not
    LD_file_path <- find_valid_file_path(LD_meta_file_path, LD_file_path)
    bim_file_path <- find_valid_file_path(LD_meta_file_path, bim_file_path)
```

this is now redundant.

R/LD.R
gaow: How about this:

```r
combined_LD_variants <- lapply(LD_matrices_list, function(x) x$extracted_LD_variants)

create_combined_LD_matrix <- function(LD_matrices, variants) {

mergeVectors <- function(vectors) {
  merged <- vectors[[1]]
  sapply(vectors[-1], function(v) { if (tail(merged, 1) == v[1]) merged <<- c(merged, v[-1]) else merged <<- c(merged, v) })
  merged
}
    
unique_variants <- mergeVectors(variants)  

```

R/LD.R
gaow: if you use my `mergeVectors` above, it should work and you can remove this line?

gaow:
Also since you remove `find_valid_file_path` from `process_LD_matrix <- function(LD_meta_file_path, LD_file_path, bim_file_path)` then this function no longer needs `LD_meta_file_path` please remove that too.

PR #89, opened 2024-01-30
add post processing for coloc res

rfeng2023:


gaow:
@rfeng2023 thanks it looks good overall although i htink some details we need to finalize based on simulation results. Before i look it more carefully i remeber you mentioned to me earlier that there are some change you will still make to it --- is it ready to review & merge or still work in progress?

rfeng2023:
Hi @gaow , I reformatted input file to put the first layer back, and also changed the way to grep analysis region (although if Susie's result interface change as discussed today, the current code can still be made simpler). Please go ahead to review it. Thanks!

R/encoloc.R
```diff
@@ -68,6 +68,123 @@ xqtl_enrichment_wrapper <- function(xqtl_files, gwas_files,
...
+#' @importFrom susieR get_purity
+# Function to calculate purity
+calculate_purity <- function(variants, ext_ld, squared) {
+  # This is a placeholder for calculating purity, adjust as per your actual function
+  purity <- matrix(get_purity(variants, Xcorr = ext_ld, squared), 1, 3)
+  purity
+}
+
+# Main processing function
+process_coloc_results <- function(coloc_result, LD_meta_file_path,analysis_script_obj, PPH4_thres = 0.8, coloc_pip_thres = 0.95, squared = FALSE, min_abs_corr = 0.5, null_index = 0, coloc_index = "PP.H4.abf", analysis_region) {
```
gaow, line 115:
```r
      fitted[[r]]$region_info = list(region=parse_region("${_meta_info[0]}"), name="${_meta_info[2]}")
```

I add the above to https://github.com/cumc/xqtl-pipeline/commit/cd5398fed292fb87ffe82d33314f647d1c7aab38 so you don't need to parse from script loaded

R/encoloc.R
```diff
@@ -68,6 +68,123 @@ xqtl_enrichment_wrapper <- function(xqtl_files, gwas_files,
...
+# Function to calculate cumulative sum
+calculate_cumsum <- function(coloc_results) {
+  cumsum(coloc_results[, 2])
+}
+
+# Function to load and extract LD matrix
+load_and_extract_ld_matrix <- function(ld_meta_file_path, analysis_region, variants) {
+  # This is a placeholder for loading LD matrix, adjust as per your actual function
+  ld_ref <- load_LD_matrix(LD_meta_file_path = ld_meta_file_path, region = analysis_region)
+  ext_ld <- ld_ref[[2]][variants, variants]
```
gaow, line 102: Instead of `[[2]]` I would use `$` followed by the exact name of the object

R/encoloc.R
```diff
@@ -68,6 +68,123 @@ xqtl_enrichment_wrapper <- function(xqtl_files, gwas_files,
...
+  ext_ld
+}
+
+#' @importFrom susieR get_purity
+# Function to calculate purity
+calculate_purity <- function(variants, ext_ld, squared) {
```
gaow, line 108: For all this and other functions that are internal, please use 

```
#' @noRd
```

R/encoloc.R
gaow: Also regarding purity -- in my branch of susieR I also use median abs corr, https://github.com/gaow/susieR/blob/master/R/susie_utils.R#L328 that you can also consider?

rfeng2023: Hi @gaow , in your version I found that you don't consider squared anymore, do we still keep that choice? And also you did some modifications on get purity. If we decide to use the new version, I will replace this whole part. Thanks!

gaow: @rfeng2023 i don't think we ever considered squared correlation. It is always abs correlation either the min or median. You can use my new version to be consistent with our susieR analysis.

rfeng2023: @gaow okay. will do

R/encoloc.R
```diff
@@ -99,11 +99,32 @@ calculate_cumsum <- function(coloc_results) {
 load_and_extract_ld_matrix <- function(ld_meta_file_path, analysis_region, variants) {
   # This is a placeholder for loading LD matrix, adjust as per your actual function
   ld_ref <- load_LD_matrix(LD_meta_file_path = ld_meta_file_path, region = analysis_region)
-  ext_ld <- ld_ref[[2]][variants, variants]
+  ext_ld <- ld_ref$combined_LD_matrix[variants, variants]
   ext_ld
 }
 
-#' @importFrom susieR get_purity
+# Subsample and compute min, mean, median and max abs corr.
```
gaow: Sorry you can still use susieR::get_purity ... it is just that this susieR is from gaow/susieR which is. the case in the pecotmr image already. There is no need to redefine it. 

At some point of development I'm going to merge gaow/susieR to stephenslab/susieR.

R/encoloc.R
```diff
@@ -112,7 +133,7 @@ calculate_purity <- function(variants, ext_ld, squared) {
 }
 
 # Main processing function
-process_coloc_results <- function(coloc_result, LD_meta_file_path,analysis_script_obj, PPH4_thres = 0.8, coloc_pip_thres = 0.95, squared = FALSE, min_abs_corr = 0.5, null_index = 0, coloc_index = "PP.H4.abf", analysis_region) {
+process_coloc_results <- function(coloc_result, LD_meta_file_path,analysis_script_obj, PPH4_thres = 0.8, coloc_pip_thres = 0.95, squared = FALSE, min_abs_corr = 0.5, null_index = 0, coloc_index = "PP.H4.abf", analysis_region, median_abs_corr = NULL) {
```
gaow, line 143: this is okay for the package so nothing to change it here setting it to NULL by default. in practice when we use this function, we can consider using median_abs_corr the same as what i did in `susie_twas` pipeline inside of cis_workhorse?

gaow: Actually you will introduce a bug in:

```
is_pure = which(purity[,1] >= min_abs_corr | purity[,2] >= median_abs_corr)
```

this line will not work if median_abs_corr is NULL

PR #95, opened 2024-02-01
Update on allele_qc and susie_wrapper

SamuelSun2000:


SamuelSun2000:
1. Not making the variant id as row names, because one reference id can correspond to two variant at the same position! Making it as row names will incur duplicated row names (and error), so best to make it a column.
2. Recode A1 and A2 after sign flip.
3. Changed the input format of raiss.

PR #97, opened 2024-02-01
Store the imputation and filter results

SamuelSun2000:
I want to keep the imputation result of z scores and LD matrix (which can filter out some variants), otherwise it's hard to use post processor to get other coverage


gaow:
@SamuelSun2000 it's okay by me to keep z scores but not LD matrix. We should be able to load LD matrix when you need it on the fly?

gaow:
as long as you keep the variant names of your z-scores

SamuelSun2000:
Make sense. I recalled that there are variant names remaining there.

PR #103, opened 2024-02-04
Change the column names so that allele_qc and raiss can fit

SamuelSun2000:
Also, the output will include imputation result of RAISS


R/allele_qc.R
```diff
@@ -36,7 +36,7 @@ convert_to_dataframe <- function(variant_id) {
   }
 
   # Check if id1 is in the first vector format
-  if (any(grepl(":", variant_id[1])) && any(grepl("_", variant_id[1]))) {
+  if (any(grepl(":", varaint_id[1])) && any(grepl("_", variant_id[1]))) {
```
gaow, line 39: @SamuelSun2000 did you test the code. Obviously this is a typo `varaint_id`

PR #111, opened 2024-02-05
Change test content 

SamuelSun2000:
Unit tests passed


SamuelSun2000:
The logic is:

In susie_rss_qc() function, if we need imputation, it will run 2 susie_rss. first susie_rss with ZR_discrep = T is for detection of outlier.
The second susie_rss will run with imputed z scores.
However, if we need conditional analysis on imputed result. THe first round and second round susie_rss will both have max_iter = 1 as input (they all have ... ). So that The detected outlier number will be influenced, to avoid that, I specified the max_iter = 100 (default) in the first round outlier.
Let me know if there are other decent solutions!

PR #115, opened 2024-02-08
fix functions

AnqiWang2021:


R/mr.R
```diff
@@ -6,107 +6,110 @@ calc_I2 = function(Q, Est) {
...
-      merge_snp <- intersect(qtl_finemap$snp, outcome_QC$variant_allele_flip)
-      
-      format_input <- qtl_finemap[match(merge_snp, qtl_finemap$snp), ] %>% 
-        mutate(gene_id = gene_id) %>% 
-        cbind(., outcome_QC[match(merge_snp, outcome_QC$variant_allele_flip), ] %>% 
-                select(beta, se) %>% 
-                rename("bhat_y" = "beta", "sbhat_y" = "se"))
-      
-      mr_format_input <- rbind(mr_format_input, format_input)
+mr_format<- function(susie_result, condition, gwas_sumstats_db, sets = "sets", coverage = "coverage_0.95", allele_qc = TRUE) {
```
gaow, line 21: Actually @AnqiWang2021 i am not sure why you need sets because the coverage table will contain everything including all variants inside of the 0.95 CS?

PR #117, opened 2024-02-13
Update for susie_rss

SamuelSun2000:
1. In LD.R have a ref_panel which is a dataframe contain chrom, pos, A1 A2 and variant_id. THis is convenient for susie_rss downstream pipeline.
2. Change the column name from Z to z so that we don't need to change it in multilple functions.
3. add three wrapper function to serve in susie_rss pipeline.


R/raiss.R
```diff
@@ -3,11 +3,11 @@
 #' This function is a part of the statistical library for SNP imputation from:
 #' https://gitlab.pasteur.fr/statistical-genetics/raiss/-/blob/master/raiss/stat_models.py
 #' It is R implementation of the imputation model described in the paper by Bogdan Pasaniuc,
-#' Noah Zaitlen, et al., titled "Fast and accurate imputation of summary
+#' Noah zaitlen, et al., titled "Fast and accurate imputation of summary
```
gaow, line 6: Should still be Zaitlen .. .

R/susie_wrapper.R
```diff
@@ -125,6 +121,113 @@ susie_rss_wrapper <- function(z, R, bhat, shat, n = NULL, var_y = NULL, L = 10,
...
+    single_effect_res = susie_rss_wrapper(z = z, R = LD_extract, bhat = NULL, shat = NULL, L = 1, n = n, var_y = var_y)
+    if (!any(single_effect_res$pip > 0.05)) {
+        message("No variant has a PIP value greater than 0.05 for single effect model. Will return single effect model result.")
+        final_result$single_effect_result = single_effect_res
+    } else {  
+        # 1. Output 1 & 2 (default):  single effect result + susie_rss no QC
+        result_noqc = susie_rss_wrapper(z = z, R = LD_extract, bhat = NULL, shat = NULL, n = n, L = L, var_y = var_y)
+        result_noqc_post = susie_post_processor(result_noqc, data_x = LD_extract, data_y = list(z), mode = "susie_rss")
+        final_result$noqc_result = result_noqc_post
+        final_result$single_effect_result = single_effect_res
```
gaow, line 206: we also need single effect result post- processed so we get different coverages.

R/susie_wrapper.R
```diff
@@ -125,6 +121,113 @@ susie_rss_wrapper <- function(z, R, bhat, shat, n = NULL, var_y = NULL, L = 10,
...
+        if(QC){
+            result_qced = susie_rss_qc(z = z, ref_panel = ref_panel, R = LD_extract, n = n, L = L, impute = impute_QC,
+            rcond = rcond, R2_threshold = R2_threshold, max_L = max_L, l_step = l_step, var_y = var_y)
+            var_impute_kept = names(result_qced$qc_impute_result$pip)
+            result_qced_impute_post = susie_post_processor(result_qced$qc_impute_result, data_x = LD_extract[var_impute_kept, var_impute_kept, drop = FALSE], data_y = list(z = result_qced$qc_impute_result$z), mode = "susie_rss")
+            final_result$qc_only_result = result_qced$qc_only_result
```
gaow, line 212: for qc only result we also need post-processed. Be consistent with the other analysis

R/susie_wrapper.R
```diff
@@ -125,6 +121,113 @@ susie_rss_wrapper <- function(z, R, bhat, shat, n = NULL, var_y = NULL, L = 10,
...
+            final_result$qc_impute_result = result_qced_impute_post
+            result_qced$impute_table$chrom = as.numeric(result_qced$impute_table$chrom)
+            final_result$qc_impute_table = result_qced$qc_impute_table
+        }
+        if(cond_analysis){
+            conditional_noqc = susie_rss_wrapper(z = z, R = LD_extract, bhat = NULL, shat = NULL, n = n, L = L, max_iter = 1, var_y = var_y)
+            conditional_qced = susie_rss_qc(z = z, ref_panel = ref_panel, R = LD_extract, n = n, L = L, impute = impute_QC, var_y = var_y,
+            rcond = rcond, R2_threshold = R2_threshold, max_L = max_L, l_step = l_step, max_iter = 1)
+            final_result$cond_noqc = conditional_noqc
+            final_result$cond_qced = conditional_qced
```
gaow, line 222: For all of these results we need post-processed, again, for different coverages and the top_loci table. That is the whole point of summarizing the results in a concise and informative way to the best of our knowledge.

R/susie_wrapper.R
```diff
@@ -125,6 +121,114 @@ susie_rss_wrapper <- function(z, R, bhat, shat, n = NULL, var_y = NULL, L = 10,
...
+    return(sumstat_processed)
+    
+}
+
+
+susie_rss_pipeline = function(sumstat, R, ref_panel, bhat = NULL, shat = NULL, n, L, var_y, QC = TRUE, impute = TRUE, cond_analysis = TRUE, rcond, R2_threshold, max_L, l_step, minimum_ld){
+        final_result = list()
+        z = sumstat$z
+        LD_extract = R[sumstat$variant_id, sumstat$variant_id, drop = FALSE]
+        single_effect_res = susie_rss_wrapper(z = z, R = LD_extract, bhat = NULL, shat = NULL, L = 1, n = n, var_y = var_y)
```
gaow, line 188: why is bhat and shat NULL here? why not use the input from the function?

R/susie_wrapper.R
```diff
@@ -125,6 +121,114 @@ susie_rss_wrapper <- function(z, R, bhat, shat, n = NULL, var_y = NULL, L = 10,
...
+        single_effect_post = susie_post_processor(single_effect_res, data_x = LD_extract, data_y = list(z), mode = "susie_rss")
+        final_result$single_effect_result = single_effect_post
+    
+        result_noqc = susie_rss_wrapper(z = z, R = LD_extract, bhat = NULL, shat = NULL, n = n, L = L, var_y = var_y)
```
gaow, line 192: bhat sbhat should not be NULL here.

R/susie_wrapper.R
```diff
@@ -125,6 +121,114 @@ susie_rss_wrapper <- function(z, R, bhat, shat, n = NULL, var_y = NULL, L = 10,
...
+            final_result$qc_impute_result = result_qced_impute_post
+            final_result$qc_only_result = result_qced_only_post
+            
+            result_qced$qc_impute_filter_table$chrom = as.numeric(result_qced$qc_impute_filter_table$chrom)
+            result_qced$qc_impute_nofilter_table$chrom = as.numeric(result_qced$qc_impute_nofilter_table$chrom)
+            final_result$qc_impute_filter_table = result_qced$qc_impute_filter_table
+            final_result$qc_impute_nofilter_table = result_qced$qc_impute_nofilter_table
+        }
+
+        if(cond_analysis){
```
gaow, line 215: let's use the parameter name `bayesian_conditional_analysis` instead of `cond_analysis`

R/susie_wrapper.R
```diff
@@ -125,6 +121,114 @@ susie_rss_wrapper <- function(z, R, bhat, shat, n = NULL, var_y = NULL, L = 10,
...
+            conditional_noqc = susie_rss_wrapper(z = z, R = LD_extract, bhat = NULL, shat = NULL, n = n, L = L, max_iter = 1, var_y = var_y)
```
gaow, line 216: bhat, and sbhat fix these.

R/susie_wrapper.R
```diff
@@ -125,6 +121,114 @@ susie_rss_wrapper <- function(z, R, bhat, shat, n = NULL, var_y = NULL, L = 10,
...
+            conditional_noqc_post = susie_post_processor(conditional_noqc, data_x = LD_extract, data_y = list(z = conditional_noqc$z), mode = "susie_rss")
```
gaow, line 217: Also for conditional analysis, let's do it simpler. I'll comment it on slack.

SamuelSun2000:
Debugged

PR #121, opened 2024-02-18
Add tests for file_utils.R

Travyse:
Added some tests for file_utils.R. I don't think I overwrote the most recent merge, but if I did I can re-push the changes.


gaow:
Thanks @Travyse no worries, recent pushes does not have any changes for these file I/O. We are putting together some pipelines into pecotmr for various analysis.

PR #129, opened 2024-02-26
Use furrr

danielnachun:
This PR replaces `doParallel` and `foreach` with `furrr`.  In local testing, I still get an error for the parallel `twas_weights` test - let's see if that also occurs in CI.


gaow:
@danielnachun please merge it as you see fit for yourself. Looks good to me. Thanks!

danielnachun:
Thanks, everything should be good to for removing `foreach`.  It looks like `doMC` is used internally by `cv.glmnet`, so for now we can't remove it.

PR #131, opened 2024-02-28
Change A1 and A2

SamuelSun2000:


R/LD.R
```diff
@@ -174,7 +174,7 @@ process_LD_matrix <- function(LD_file_path, bim_file_path) {
     # Process variant names from file paths
     LD_variants <- bim_file_name %>%
               read.table(.) %>%
-              setNames(c("chrom","variants","GD","pos","A2","A1"))%>%
+              setNames(c("chrom","variants","GD","pos","A1","A2"))%>%
```
gaow, line 177: are you sure this line change is correct? A2 should be position 3, A1 position 4?

gaow: When Tosin make the bim file I think it is also done by this order, see his code on this page:

https://cumc.github.io/xqtl-pipeline/code/reference_data/ld_reference_generation.html

the part that bim file are written

R/susie_wrapper.R
```diff
@@ -342,7 +342,7 @@ susie_rss_pipeline = function(sumstat, R, ref_panel, n, L, var_y, QC = TRUE, imp
 #'
 #' @param z Numeric vector of z-scores corresponding to the effect size estimates, with names matching the reference panel's variant IDs.
 #' @param R Numeric matrix representing the LD (linkage disequilibrium) matrix.
-#' @param ref_panel Data frame with 'chrom', 'pos', 'variant_id', 'A0', 'A1' column that matches the names of z.
+#' @param ref_panel Data frame with 'chrom', 'pos', 'variant_id', 'A1', 'A2' column that matches the names of z.
```
gaow, line 345: A1, A2 or A2, A1?

gaow:
@SamuelSun2000 please address my review comments. Thanks!

R/LD.R
SamuelSun2000: I checked it and also searched in plink documentation:
image

so I think my update here is correct

R/susie_wrapper.R
SamuelSun2000: THe order in this table (reference panel) will not change the result, so A1 A2 / A2 A1 both works. About the order in the output I prefer A1 A2, since a lot sumstat file put the order of effect allele then other allele

R/LD.R
SamuelSun2000: A1 is alt (effect allele) and A2 is ref (other allele). So A1 in front of A2. In Tosin's code it's correct: alt in front of ref, which means A1 in front of A2

PR #139, opened 2024-03-01
update mash_wrapper.R

AnqiWang2021:


gaow:
@danielnachun i think codecov is now being pick about all lines added should be tested, otherwise it will not pass? We should relax it I think. I'm going to merge it for now.

PR #144, opened 2024-03-05
mvsusie_mrmash pipeline modifications

Chunmingl:
pipeline examples details were submitted in the PR


R/misc.R
```diff
@@ -458,7 +458,9 @@ load_regional_multivariate_data <- function(matrix_y_min_complete = NULL, # when
   }
   return (list(
         residual_Y = Y,
+        residual_X = dat$residual_X, #added
```
gaow, line 461: I don't think we need these in your mnm_pipeline

R/misc.R
```diff
@@ -458,7 +458,9 @@ load_regional_multivariate_data <- function(matrix_y_min_complete = NULL, # when
...
         residual_Y_scalar = Y_scalar,
+        residual_X_scalar = dat$residual_X_scalar,#added
```
gaow, line 463: This might be needed if scale is TRUE but you would need scaling factor for the X matrix not the residual of X, in your mnm_pipeline, so this is incorrect anyways. 

I'm not sure if we actually kept track of scaling factor for the `X` matrix, though ... if we have not and you need that, you should compute and keep it and add it like `$X_scalar` but not `$residual_X_scalar`

R/twas.R
```diff
@@ -452,6 +456,12 @@ mrmash_weights <- function(...) {
     res <- mrmash_wrapper(...)
     return(coef.mr.mash(res)[-1,])
 }
+                   
+#' @importFrom mvsusieR coef.mvsusie
+#' @export               
+mvsusie_weights <- function(mvsusie_fit=NULL,...) {
```
gaow, line 462: This is also wrong ... when mvsusie_fit is null we need to fit mvsusie. Just like this function:

https://github.com/cumc/pecotmr/blob/main/R/twas.R#L390

Otherwise your function by default would not work

R/multivariate_pipeline.R
```diff
@@ -0,0 +1,219 @@
+
+# FIXME: use a single prior for input into mvsusie and mr.mash
```
gaow, line 2: @Chunmingl are these FIXME still relevant?

Chunmingl:
Would it be better if I close this one and open a new freash PR ?

PR #153, opened 2024-03-11
Add unit tests for twas_scan and encoloc

Travyse:
- Increase coverage for the `twas_scan` and `encoloc` scripts. 
- The change I made in the `encoloc.R` script is just in the arrangement of variables in the function definition. It's a little awkward to have parameters with no defaults come after parameters with default values
- Local tests show there's a failed test for mash_wrapper that I can look into, but I want to push what I have so far


gaow:
Thanks @Travyse . 

> Local tests show there's a failed test for mash_wrapper that I can look into, but I want to push what I have 

Anqi will fix it next week. Don't worry

PR #154, opened 2024-03-11
fix

AnqiWang2021:


R/mash_wrapper.R
```diff
@@ -389,18 +387,17 @@ merge_susie_cs <- function(susie_fit, coverage = "cs_coverage_0.95", complementa
...
 load_multitrait_R_sumstat <- function(
     susie_fit, sumstats_db, coverage = NULL, top_loci = FALSE,
     filter_file = NULL, remove_any_missing = TRUE, max_rows_selected = 300, nan_remove = FALSE,
     exclude_condition, complementary = FALSE) {
   extract_data <- function(sumstats_db) {
-    bhat <- as.data.table(cbind(sumstats_db$variant_names, sumstats_db$sumstats$betahat))
-    sbhat <- as.data.table(cbind(sumstats_db$variant_names, sumstats_db$sumstats$sebetahat))
-    setnames(bhat, c("variants", "bhat"))
-    setnames(sbhat, c("variants", "sbhat"))
+    bhat <- data.table::as.data.table(cbind(sumstats_db$variant_names, sumstats_db$sumstats$betahat))
```
gaow, line 397: this is not the right way to use functions in packages. Why not `@importFrom data.table::as data.table`? Or rather why not `as.data.frame`?

PR #162, opened 2024-03-19
Fix sdpr on Apple Silicon

danielnachun:
SDPR as it was originally implemented uses x86 intrinsics to speed up MCMC sampling.  To make this portalbe to Apple Silicon, we instead use the headers from SIMDEverywhere, which can alias x86 intrinsics to their ARM NEON equivalents.  This allows for us to get the full benefit of architecture specific intrinsics without having to rewrite any assembly code.


.github/recipe/meta.yaml
```diff
@@ -24,7 +24,7 @@ requirements:
   host:
     - bioconductor-qvalue
     - r-bedtoolsr
-    - r-base>=4.2
+    - r-base==4.3.3
```
gaow, line 27: Should this be >=4.3.3?

gaow:
Thanks @danielnachun my only other comment in addition to R version is that do we want to remove the  arm, mips and wasm folders if the only folder  used is x86?

.github/recipe/meta.yaml
danielnachun: No we want to pin the host version.  The run time version can be unbounded on the top, but if we don't pin it here, we won't get a consistent version of `r-base` when we build the package, which could give inconsistent results.  We will want to manually raise this from time to time.

danielnachun:
> Thanks @danielnachun my only other comment in addition to R version is that do we want to remove the arm, mips and wasm folders if the only folder used is x86?

I'd rather leave the SIMDE package intact in the folder just to avoid any confusion over where the code comes from.  That will make it easier to periodically update it.  Eventually I will try to find a way to use an external copy of SIMDE instead, but it doesn't provide a `pkg-config` file so that makes it a bit trickier to find automatically.

danielnachun:
Yes, I took the amalgamated release from https://github.com/simd-everywhere/simde/releases/download/v0.8.0/simde-amalgamated-0.8.0.tar.xz and just placed it in the `src` folder.  Let me try deleting the `mips` and `wasm` folders because those aren't going to be used by ARM.  We do need to keep the ARM folder because that contains the aliases to help translate to ARM intrinsics from x86.

danielnachun:
Okay deleting those folders didn't cause any issues.  I think this is ready to merge.  There are some failing unit tests but these were already known and are being fixed as far as I know

PR #165, opened 2024-03-21
replace pi_gwas with num_gwas

rfeng2023:


gaow:
@rfeng2023 the related unit test is broken. Please fix when you get a chance, thanks!

PR #168, opened 2024-03-26
variant selection for ctwas

Chunmingl:
#### edits on `allele_qc.R`:
- allele_qc(): The step of `variant_id_to_df(target_data)` was not able to apply to weight matrices as the `target data`, as variant_id_to_df is expecting variant_id type of input with chrom, pos, A1, A2 with delimiters. 
- `target data` seems to be a data frame that provides the column indexes that intend to be flipped, which indices were identified by the first two parameter input as `target_variants`, and `ref_variants`.
#### edits on  `misc.R`
- load_ctwas_weights(): drafted based on `load_twas_weights` function, as the original function reads in weights already, however, the additional selection and evaluation process is based on the `susie_twas` twas_cv information which is not available from `load_twas_weights` output. So this function does similar steps, but additionally performs variant selection looping through cs, on imputable gene_condition, and outputs the model selection and variant selection information.
#### edits on `susie_wrapper.R`
- adjust_susie_weights() also performs allele_qc() step with weight matrices as the target data.


gaow:
@Chunmingl as discussed please send in a new one with a bit more clean-up, thanks!

Chunmingl:
Ok sounds good! Will send it today

PR #170, opened 2024-03-26
Added bayes_X_weights functions to regularized regression

lambdr:
added a general bayes_alphabet_weights wrapper, and then specific wrappers for each prior


R/regularized_regression.R
```diff
@@ -409,3 +409,74 @@ mrash_weights <- function(X, y, init_prior_sd = TRUE, ...) {
...
+#' set2 <- sample(1:ncol(W), 5)
+#' sets <- list(set1, set2)
+#' g <- rowSums(W[, c(set1, set2)])
+#' e <- rnorm(nrow(W), mean=0, sd=1)
+#' y <- g + e
+#' bayes_l_weights(y=y, X=X, Z=Z)
+#' bayes_r_weights(y=y, X=X, Z=Z)
+#' @importFrom qgg gbayes
+#' @export
+bayes_alphabet_weights <- function(y, X, method, Z=NULL){
```
gaow, line 438: Our interface https://github.com/cumc/pecotmr/blob/main/R/regularized_regression.R#L384 the input should be X, y not y, X

R/regularized_regression.R
```diff
@@ -409,3 +409,74 @@ mrash_weights <- function(X, y, init_prior_sd = TRUE, ...) {
...
+  # check for identical row lengths of response and genotype
+  if (!(length(y) == nrow(W))){
```
gaow, line 440: I don't see `W` on your input. Should it be `Z` or X? Again X is genotype data, Z is covariates.

R/regularized_regression.R
```diff
@@ -409,3 +409,74 @@ mrash_weights <- function(X, y, init_prior_sd = TRUE, ...) {
...
+        stop("Genotype and covariate matrices must have same number of rows")
+      }
+    }
+  
+  model = gbayes(y=y, 
+                 W=X,
+                 X=Z,
+                 method=method)
+  
+  return(model$bm)
```
gaow, line 455: Double check that the return contains the same number of coef as you run other existing methods such as lasso -- i worry that it may or may not contain the intercept. Just make sure they are consistentn

gaow:
@lambdr please address my comments in your updated commit. Thanks!

R/regularized_regression.R
lambdr: All functions have been updated to X, y format.

R/regularized_regression.R
lambdr: This error has been fixed

R/regularized_regression.R
lambdr: Yes, the returned vector has the same number of betas as there are snps

gaow:
@lambdr good work!

PR #172, opened 2024-03-26
A intermediate update

SamuelSun2000:


R/sumstats_qc.R
```diff
@@ -52,8 +52,10 @@ rss_basic_qc <- function(sumstats, LD_data, skip_region = NULL) {
     allele_flip$target_data_qced <- allele_flip$target_data_qced %>%
       filter(!(variant_id %in% skip_variant))
   }
-
-  sumstats_processed <- allele_flip$target_data_qced %>% arrange(pos)
+  # remove totally same variant names, which is common to see in lifted data.
```
gaow, line 55: I dont understand this --- the output of allele_qc would have duplicates? How does it happen in the first place? Would it make more sense to process it right in the allele_qc function so anything else using allele_qc would not have this issue? 

Still I don't get why it happens in the first place. Sounds like something unusual and should be addressed at its root not by removing it after.

gaow:
@SamuelSun2000 i had one comment above regarding duplicated variants please address that. Also for this file, 

inst/prototype/Dentist_R_cpp_comparison.ipynb are you sure about the changes? Just make sure that Rui is okay with it. I can see your version contain more information but also removed something that seems good to have, from Rui's previous version.

R/sumstats_qc.R
SamuelSun2000: This case is indeed unusual, the phenomenom is found in Thomas' case. 

What I found is after (the original version of) allele qc, there are some variants having exactly same variant name: chr, pos and alleles, but **different z scores**. Then it will lead to some bugs since we cannot decide on which z to use, and allele qc cannot effectively remove these variants since it's checking totally the same rows.  Thomas said it's due to the lift from hg19 to hg38, after mapping some different variants can end up in the same position, and if happen to have same alleles, which can be due to the higly duplicated region, will lead to results like this.

My suggestion to him is simply remove these variants that I think is reasonable to do so. But I agree, we can put this line in allele QC?

SamuelSun2000:
@gaow I've responded to the comment above. As for the Dentist_R_cpp_comparison, I'm sure Rui haven't pushed the final version that is doing exact match for every steps. We will make a final wrap up later including input from both our ends.

Also I'd like to address a logic in the change of code: I put allele qc before checking if any pip derived from simple linera regression larger than cutoff. QC don't cost too much, and only after this can I match the variants in sumstat with LD matrix, so that susie_rss can accept them and run.

R/sumstats_qc.R
gaow: Thanks @SamuelSun2000 this is clear to me know.  in that case in our alele QC we should detect the situation you decribe above, and throw a proper error message to prompt the users to make changes. Let's not just remove duplicates for them. They need to decide what to keep.

R/raiss.R
```diff
@@ -149,6 +149,8 @@ merge_raiss_df <- function(raiss_df, known_zscores) {
   # Remove the extra columns resulted from the merge (e.g., z.x, z.y)
   merged_df <- merged_df[, !colnames(merged_df) %in% c("z.x", "z.y")]
   merged_df <- arrange(merged_df, pos)
+  # remove beta, se to avoid confusion, since they are not imputed, we only impute z
+  merged_df = merged_df[, !colnames(merged_df) %in% c("beta", "se", "pvalue")]
```
gaow, line 153: Do you think it is helpful if instead of removing we set them to NA? Or you think it does not matter? Remeber we want to export our QC+imputed data to text file for other analysis. So i think providing as complete information as possible is good idea?

PR #189, opened 2024-04-11
Allows handling lifted over LD for RSS

hsun3163:
1. Properly filter out the chrom = 0 in a LD block, duplicates of position in a LDblock, and the snp that are lifted out of the LD block. This is done at this point instead of before the analysis because the chr0 needs to be retained such that they have the same dimension of the xz file 

2. Add the option to change the allele_qc setting to remove indel (still kept as false by default.).

Both changes have be tested to successfully fine-map the gwas using the lifted LD, how valuable the lifted LD is is another question though.

The changes regarding eQTL have not be implemented and  tested yet.


hsun3163:
Now I added an accomodation of eQTL. basically just retains only the rows in a sumstat file that match a certain pattern. This function are only invoked by setting eQTL = TRUE in the sos notebook. By default the pattern = "" so all the gwas data can be gathered intact.

R/LD.R
```diff
@@ -199,11 +200,16 @@ process_LD_matrix <- function(LD_file_path, bim_file_path) {
     # If the matrix is lower diagonal, transpose the lower triangle to the upper triangle
     LD_matrix[upper.tri(LD_matrix)] <- t(LD_matrix)[upper.tri(LD_matrix)]
   }
+
+    ### remove the variants that is zero (indicated not lifted) or not in the most frequent chrom in the mode, this rely on the fact that majority of the chromosome is lifted correctly. 
+    ### Remove all the variants whose pos are duplicated, for they are ambigouse
+  mode_chrom <- LD_variants %>% filter(chrom != 0) %>% pull(chrom) %>% table() %>% which.max() %>% names()    
+  LD_variants <- LD_variants %>% filter(chrom ==  mode_chrom)%>%distinct(pos,.keep_all = T)
```
gaow, line 207: we shoud not dedup here.

R/file_utils.R
```diff
@@ -696,61 +696,84 @@ load_twas_weights <- function(weight_db_files, conditions = NULL,
 #' @return A list of rss_input, including the column-name-formatted summary statistics,
 #' sample size (n), and var_y.
 #'
+#' @importFrom dplyr mutate group_by summarise %>%
```
gaow, line 699: use magrittr for this %>%

gaow:
@hsun3163 you have many components in this PR that i can see some useful but the majority needs rework into the new interface we talked about that is more generic and efficient (tabix). I'm tehrefore closing this PR. In your next iteration could you please also take what is still useful in this PR and add to the next PR?

PR #197, opened 2024-05-06
summary stat functions 

lambdr:
I've added a modified version of the qgg::gbayes() function that works with summary statistics (LD matrix, beta estimates, and std errors of betas), as well as wrappers that run this function with specified priors, e.g. bayes_n_rss_weights and bayes_r_rss_weights.

Additionally, I've added wrappers for the mr_ash_rss and prs_cs functions that return the estimated weights.

### DIFF
--- a/R/regularized_regression.R
+++ b/R/regularized_regression.R
@@ -581,7 +581,8 @@ bayes_r_weights <- function(X, y, Z = NULL) {
 #' \item{ve}{mean residual variance}
 #' \item{vg}{mean genomic variance}
 #'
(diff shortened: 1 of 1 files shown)


lambdr:
By default, specification of BayesR as the method forces the mixing proportions of the normal distributions to be (0.95, 0.02, 0.02, 0.01). If it is important that the user be able to specify the mixing proportions, that change can be made.

danielnachun:
Can you rename your remote branch to something other than `main`?  This can create conflicts downstream.

PR #198, opened 2024-05-06
customized functions for cTWAS multigroup

Chunmingl:
- `ctwas_wrapper.R` functions are tested by installing from local gitrepository and works with pipeline named as `ctwas_input_processing`. 
 - Some older functions were removed as they may no longer be relevant for current version of ctwas multigroup 

- `LD.R` edit is for extract and then keep variance information from LD for ctwas analysis, this modification works for both the newly processed LD (April), and is conditioned to work the same for the older LD (Jan).
- `twas.R` edit is for keeping argument terms of susie_fit in the same dimension of variants with the `X_filtered` at the internal function of `compute_method_weights`.
- `regularized_regression.R` set default parameter for nit=5000, nburn=1000, nthin=5


gaow:
@Chunmingl to finalize on it, i thikn you need to move `inst/ctwas_wrapper.R` into the `R` folder, install and test it locally with MWE. Then start to incorporate it to the pipeline on xqtl-protocol

R/twas.R
```diff
@@ -464,11 +464,23 @@ twas_weights <- function(X, Y, weight_methods, num_threads = 1, seed = NULL) {
...
     args <- weight_methods[[method_name]]
-
+    
     # Remove columns with zero standard error
     valid_columns <- apply(X, 2, function(col) sd(col) != 0)
-    X_filtered <- X[, valid_columns]
-
+    X_filtered <- as.matrix(X[, valid_columns])
+                           
+    if (method_name=="susie_weights"){
```
gaow, line 472: This is good thinking but I don't think we should handle it for users here. We should throw an error message. That is, in the `susie_weight` function if the number of X columns does not match the dimension of the fitted model, we throw an error. Woudl you implement that instead and remove your change here?

Chunmingl: @gaow This has been addressed

PR #203, opened 2024-05-28
Add parameter "gene" 

SamuelSun2000:
so that we can target one specific gene in the sumstats


R/file_utils.R
```diff
@@ -700,9 +700,13 @@ load_twas_weights <- function(weight_db_files, conditions = NULL,
 #'
 #' @importFrom data.table fread
 #' @export
-load_rss_data <- function(sumstat_path, column_file_path, n_sample = 0, n_case = 0, n_control = 0) {
+load_rss_data <- function(sumstat_path, column_file_path, gene = NULL, n_sample = 0, n_case = 0, n_control = 0) {
```
gaow, line 703: I would like to be a bit generic -- can we call this "region" not "gene"?

gaow: Or even more generically, 

```
function(sumstat_path, column_file_path, filter = NULL, n_sample = 0, n_case = 0, n_control = 0)
```

then your `filter` should be `c("column_name", "value")`

PR #205, opened 2024-06-04
update dimension mismatch, add output for method selection - work in progress

Chunmingl:


R/ctwas_wrapper.R
```diff
@@ -165,123 +147,60 @@ select_ctwas_weights <- function(weight_db_files, conditions = NULL,
...
+  # select variants 
+  names(contexts) <- rep("non_imputable", length(contexts))
+  names(contexts)[which(contexts %in% imputable_contexts)] <- "imputable"
+  ctwas_select_result <- ctwas_select(twas_data_combined, contexts, names(contexts), max_var_selection, variable_name_obj, susie_obj)
+  
   # return results
   return(list(
     susie_results = ctwas_select_result, model_selection = model_selection,
-    weights = weights, region_info = region_info
+    weights = weights, region_info=twas_data_combined$susie_results[[1]]$region_info
```
gaow, line 197: I think `region_info` is in the original weight database. It's a bit confusing to read it as `susie_result` because we don't always have to run `susie` at all to get the original weight DB.

Chunmingl: yes, indeed, I can replace that as gene=gene, the main purpose is just to identify which gene this rds is for.

PR #216, opened 2024-07-11
update enrich vignettes

rfeng2023:


gaow:
@rfeng2023 this is not what I was hoping for, but since it is so different from what i hoped, I merged it anyways.

Basically we need to start a separate page, write it clearly following from some structure on narratives (not necessarily [this](https://wanggroup.org/orientation/reproducible-research.html#suggested-format-of-lab-notebooks) which is good for daily notebooks not vignettes, but maybe something that looks closer to [this](https://stephenslab.github.io/susieR/articles/finemapping.html)) 

Just imagine, when you read this vignette as a user, do you find it both professional and informative -- that would be the "metric" of what it should look like. So when you get a chance could you start a separate vignette document and try again?

PR #223, opened 2024-07-19
integrate univ multiv twas pipeline

Chunmingl:
changes for integrate multivariate weights and univariate weights.


R/file_utils.R
```diff
@@ -574,17 +586,66 @@ load_twas_weights <- function(weight_db_files, conditions = NULL,
...
+              } else {
+                result[[name]] <- item[[name]]
+              }
+            }
+          }
+          return(result)
+        })
+    }                                                                   
+    # merge univariate and multivariate results for same gene-context pair                                                         
+    if ("mnm_rs" %in% names(combined_all_data)){
```
gaow, line 630: This needs some clean-up still. I think this can be more generic without using if ("mnm_rs" %in% names(combined_all_data)) because the task is essentailly to make it work allowing for missing variants. But we can table it.

NAMESPACE
```diff
@@ -15,6 +15,7 @@ export(bayes_n_rss_weights)
 export(bayes_n_weights)
 export(bayes_r_rss_weights)
 export(bayes_r_weights)
+export(clean_context_names)
```
gaow, line 18: This should also be an internal function. We can clean up later

PR #230, opened 2024-08-20
restore mrmash march pipeline

Chunmingl:
The following implementation were included in this PR: 

- The `null` component were checked and if found will be removed in the prior at the step of` filter_data_driven_mats ` within `mrmash_wrapper` - as the following step will add a null component to the prior matrix list. 
- `cv_seed` were removed from twas/twas_cv functions
-  unadjusted `r2` result was added in the cv performance table output
- `compute_coefficients_univ_glmnet` was disabled from running on multicore for `mrmash_wrapper`
-  `do.call` to replace `exec` in   weights_vector <- exec(method_name, X = X_filtered, y = Y[, k], !!!args) 
- `weights_tol` at the moment was removed from `mrmash_wrapper` - which we may want to further address for the threshold for it. 
- `drop=FALSE` was added to subseting of matrices in TWAS
- The weight_methods of mvsusie were dropped if `mvsusie_fitted` = `mnm_result$susie_result_trimmed` is NULL
- The first `twas_weights` if condition statement was removed as we will need to fit for mrmash and mvsusie regardless
- The default  `canonical_prior_matrices` is set to FALSE for multivariate pipeline
- The previous bug related to twas cv dimension mismatch issue was addressed


gaow:
> weights_tol at the moment was removed from mrmash_wrapper - which we may want to further address for the threshold for it.

We definitely still need this, but we might add it elsewhere because it might make more sense to use default for what's during the iterations but put in some filters on this before everythig starts. We can discuss this PM.

R/mrmash_wrapper.R
```diff
@@ -92,12 +92,15 @@ mrmash_wrapper <- function(X,
                            w0_threshold = 1e-8,
```
gaow, line 92: This threshold is still here. Is it ever used? If not, please make sure subsequent steps uses thsi parameter.

R/mrmash_wrapper.R
```diff
@@ -142,12 +139,13 @@ mrmash_wrapper <- function(X,
       hetgrid = c(0, 0.25, 0.5, 0.75, 1)
     )
     if (!is.null(data_driven_prior_matrices)) {
+      data_driven_prior_matrices <- filter_data_driven_mats(Y, data_driven_prior_matrices)
```
gaow, line 142: I think we should have some weight_tol here when filtering data driven mats.

R/mrmash_wrapper.R
```diff
@@ -193,13 +191,7 @@ compute_coefficients_glasso <- function(X, Y, standardize, nthreads, Xnew = NULL
...
 
   # Fit group-lasso
-  if (nthreads > 1) {
-    registerDoFuture()
-    plan(multicore, workers = nthreads)
-    paral <- TRUE
-  } else {
-    paral <- FALSE
-  }
+  paral <- FALSE
```
gaow, line 194: Lets not use this parameter but just set this to FALSE in cv.glmnet call

R/mrmash_wrapper.R
```diff
@@ -230,13 +222,7 @@ compute_coefficients_univ_glmnet <- function(X, Y, alpha, standardize, nthreads,
...
 
   linreg <- function(i, X, Y, alpha, standardize, nthreads, Xnew) {
-    if (nthreads > 1) {
-      registerDoFuture()
-      plan(multicore, workers = nthreads)
-      paral <- TRUE
-    } else {
-      paral <- FALSE
-    }
+    paral <- FALSE
```
gaow, line 225: same thing here -- let's remove this parameter. Also, renove the `nthreads` parameter in this context.

R/mrmash_wrapper.R
```diff
@@ -297,6 +283,7 @@ filter_data_driven_mats <- function(Y, data_driven_mats) {
 
   # Check if colnames of Y is a subset of column names of each element in data_driven_mats
   for (mat_name in names(data_driven_mats)) {
+    if (mat_name == "null") data_driven_mats[[mat_name]] <- NULL; next # remove null component
```
gaow, line 286: this is not a good check. I think you should check if all element is zero for the null component?

R/multivariate_pipeline.R
```diff
@@ -233,7 +235,7 @@ twas_multivariate_weights_pipeline <- function(
     data_driven_prior_matrices = NULL, data_driven_prior_matrices_cv = NULL, canonical_prior_matrices = FALSE, resid_Y,
     mvsusie_max_iter = 200, mrmash_max_iter = 5000,
     signal_cutoff = 0.05, coverage = 0.95, secondary_coverage = c(0.7, 0.5),
-    min_cv_maf = 0.05, max_cv_variants = -1, cv_seed = 999, cv_threads = 1, verbose = FALSE, imiss_cutoff = NULL) {
+    min_cv_maf = 0.05, max_cv_variants = -1, cv_threads = 1, verbose = FALSE, imiss_cutoff = NULL) {
```
gaow, line 238: here i guess we can also specify cv_threads? although we can make it default to 1.

R/twas.R
```diff
@@ -377,12 +380,12 @@ twas_weights_cv <- function(X, Y, fold = NULL, sample_partitions = NULL, weight_
     names(Y_pred) <- gsub("_weights", "_predicted", names(Y_pred))
 
     # Compute rsq, adj rsq, p-value, RMSE, and MAE for each method
-    # metrics_table <- matrix(NA, nrow = length(weight_methods), ncol = 5)
+    # metrics_table <- matrix(NA, nrow = length(weight_methods), ncol = 6)
```
gaow, line 383: remove this line if no longer relevant

R/mrmash_wrapper.R
fmorgante: That's the mr.mash parameter to drop components with weight smaller than that value within the algorithm. So, it should be left, I think?

R/multivariate_pipeline.R
Chunmingl: `cv_threads` is right after `cv_seed `, there is `cv_threads=1`

R/file_utils.R
```diff
@@ -536,7 +536,8 @@ load_regional_multivariate_data <- function(matrix_y_min_complete = NULL, # when
     X = X,
     maf = apply(X, 2, compute_maf),
     chrom = dat$chrom,
-    grange = dat$grange
+    grange = dat$grange,
+    X_variance= matrixStats::colVars(X)
```
gaow, line 540: @Chunmingl please use @importFrom matrixStats colVars in your roxygen strings. Dont use `::` in the code. This is an R package you need to follow the rules.

R/misc.R
```diff
@@ -125,32 +125,30 @@ filter_X <- function(X, missing_rate_thresh, maf_thresh, var_thresh = 0, Y = NUL
...
-        }
-      }
-    }
-    drop_snp_indices <- unique(drop_snp_indices)
-    if (length(drop_snp_indices)) X <- X[, -drop_snp_indices, drop = FALSE]
-    message(paste0("Dropped ", length(drop_snp_indices), " variants with condition of Y subjects, remaining ", ncol(X), "variants. "))
+  return(X)
+}
+
+# filter X and Y for TWAS analysis
```
gaow, line 131: Can you add roxygen documentation for this?

R/misc.R
```diff
@@ -167,6 +165,15 @@ filter_Y <- function(Y, n_nonmiss) {
   return(list(Y = Y, rm_rows = rm_rows))
 }
 
+filter_data_driven_priors <- function(data_driven_prior_matrices, prior_weights_min){
```
gaow, line 168: Do we have a step to also include only the context that we specify? Or are we using a different function for that?

R/mrmash_wrapper.R
```diff
@@ -291,30 +273,66 @@ compute_w0 <- function(Bhat, ncomps) {
   return(w0)
 }
 
+
 ### Filter data-driven matrices
-filter_data_driven_mats <- function(Y, data_driven_mats) {
+filter_data_driven_mats <- function(Y, data_driven_mats, data_driven_mats_w = NULL, prior_weights_min = 1e-4) {
```
gaow, line 278: Again it does not seem to make sense to me to have two functions `filter_data_driven_mats` and `filter_data_driven_priors` . Please carefully merge them.

R/mrmash_wrapper.R
```diff
@@ -291,30 +273,66 @@ compute_w0 <- function(Bhat, ncomps) {
...
+  message(paste(length(data_driven_mats), "components of data driven matrices remained after filtering. "))
+  return(list(U=data_driven_mats, w=data_driven_mats_w))
+}
 
-  data_driven_mats_filt <- lapply(data_driven_mats, function(x, to_keep) {
-    x[to_keep, to_keep]
-  }, conditions_to_keep)
 
-  return(data_driven_mats_filt)
+### Re-normalize mrmash weight w0 for total weight sum to 1
```
gaow, line 309: this is not correct roxygen format.

R/mrmash_wrapper.R
```diff
@@ -291,30 +273,66 @@ compute_w0 <- function(Bhat, ncomps) {
   return(w0)
 }
 
+
 ### Filter data-driven matrices
```
gaow, line 277: this is not correct roxygen format.

R/mrmash_wrapper.R
```diff
@@ -291,30 +273,66 @@ compute_w0 <- function(Bhat, ncomps) {
...
-  data_driven_mats_filt <- lapply(data_driven_mats, function(x, to_keep) {
-    x[to_keep, to_keep]
-  }, conditions_to_keep)
 
-  return(data_driven_mats_filt)
+### Re-normalize mrmash weight w0 for total weight sum to 1 
+#' @param w0 is the weight of mr.mash prior matrices that was generated from mr.mash() function. 
+rescale_cov_w0 <- function(w0){
+    # remove null component 
+    w0 <- w0[names(w0) != "null"]
```
gaow, line 313: Are you sure that here the null component will always have the name `null`? If not you need to use a more robust way.

R/mrmash_wrapper.R
```diff
@@ -291,30 +273,66 @@ compute_w0 <- function(Bhat, ncomps) {
...
+    
+    # split by prior group 
+    groups = sub("_[^_]+$", "",  names(w0))
```
gaow, line 316: are you sure this is a reliable and robust way to get prior groups? Possibly yes, if this is the convention that mr.mash expand grid uses so hyou are basicall reverse engineer that convention. But please double check and confirm.

R/twas.R
```diff
@@ -180,7 +179,7 @@ twas_analysis <- function(weights_matrix, gwas_sumstats_db, LD_matrix, extract_v
 #' @importFrom furrr future_map furrr_options
 #' @importFrom purrr map
 #' @export
-twas_weights_cv <- function(X, Y, fold = NULL, sample_partitions = NULL, weight_methods = NULL, seed = NULL, max_num_variants = NULL, variants_to_keep = NULL, num_threads = 1, imiss_cutoff = 1.0, min_cv_maf = 0.05, ...) {
+twas_weights_cv <- function(X, Y, fold = NULL, sample_partitions = NULL, weight_methods = NULL, max_num_variants = NULL, variants_to_keep = NULL, num_threads = 1, imiss_cutoff = 1.0, ...) {
```
gaow, line 182: are you still using `imiss_cutoff` somewhere? I dont think that is the case any more. Don't we do QC at least on imiss_cutoff for each fold of CV? i

R/twas.R
```diff
@@ -315,8 +316,9 @@ twas_weights_cv <- function(X, Y, fold = NULL, sample_partitions = NULL, weight_
       # Remove columns with zero standard error
       valid_columns <- apply(X_train, 2, function(col) sd(col) != 0)
       X_train <- X_train[, valid_columns, drop = F]
-      X_train <- filter_X(X_train, imiss_cutoff, min_cv_maf, Y = Y_train)
-      valid_columns <- colnames(X_train)
+      # X_train <- filter_X(X_train, imiss_cutoff, min_cv_maf, Y = Y_train, drop_by_Y_only = TRUE)
```
gaow, line 319: looks like you are no longer using imisscuttoff on each fold of data? i'm fine with it as long as you are fine with it. ...

gaow: and if so, please remove these commented out lines to avoid confusion

R/twas.R
```diff
@@ -328,7 +330,7 @@ twas_weights_cv <- function(X, Y, fold = NULL, sample_partitions = NULL, weight_
               args$data_driven_prior_matrices <- cv_args$data_driven_prior_matrices_cv[[j]]
             }
             if (method == "mvsusie_weights") {
-              args$prior_variance <- cv_args$data_driven_prior_matrices_cv[[j]]
+              args$prior_variance <- cv_args$mvsusie_data_driven_prior_matrices_cv[[j]]
```
gaow, line 333: Can we call it `reweighted_data_driven_prior_matrices`? Don't call it `mvsusie_data_driven_prior_matrices_cv` so we know that in this input data the weight is refined.

R/twas.R
```diff
@@ -502,7 +502,7 @@ twas_weights <- function(X, Y, weight_methods, num_threads = 1, seed = NULL) {
   if (num_cores >= 2) {
     # Set up parallel backend to use multiple cores
     plan(multicore, workers = num_cores)
-    weights_list <- names(weight_methods) %>% future_map(compute_method_weights, weight_methods, .options = furrr_options(seed = seed))
+    weights_list <- names(weight_methods) %>% future_map(compute_method_weights, weight_methods, .options = furrr_options(seed=TRUE))
```
gaow, line 505: in the previous call of `furrr_options` you have `globals` but you don't have it here? Please double check to make sure this is intended.

R/misc.R
Chunmingl: Sorry this would need to be removed and it was not used anywhere in the pipeline, I missed cleaning up this one.

R/mrmash_wrapper.R
Chunmingl: This is the function meant to be used and actually used in the multivariate pipeline.

R/mrmash_wrapper.R
Chunmingl: @fmorgante For the step of "split by prior group"  what i did was determine groups by splitting by component names from the last appearing "\_", that after removing everything after the last appearing "_", (_gridX) same names get grouped together. Would this be a safe approach for splitting these components by groups, attached are the `w0` names from mr.mash fitted using our current data driven prior.

image

R/mrmash_wrapper.R
Chunmingl: This is addressed. The roxygen documentaion for this function `filter_data_driven_mats ` is added

R/mrmash_wrapper.R
Chunmingl: I think this `null` component is introduced at the step of `S0 <- expand_covs(S0_raw, prior_grid, zeromat = TRUE)`  and we did not provide users any options to change parameters in this step, the null component should be always named as `null`. Please confirm @fmorgante

R/twas.R
Chunmingl: No, it won't be used anywhere inside of twas cv, it is now removed in the upcoming PR.

R/twas.R
Chunmingl: Yep, addressed in the new PR

R/twas.R
Chunmingl: The name is updated in the new PR

R/mrmash_wrapper.R
Chunmingl: The roxygen documentation is added for this function.

R/misc.R
Chunmingl: The roxygen documentation for `filter_X_with_Y` is added.

R/file_utils.R
Chunmingl: This is addressed in the new commit.

R/twas.R
Chunmingl: yes, this step is for twas weight computation without cv 
`twas_weights_res <- twas_weights(X = X, Y = Y, weight_methods = weight_methods)` 
future achieve it through the function 
`compute_method_weights <- function(method_name, weight_methods) `
so there are no extra parameters needed to be introduced other than weight_methods, 

The one above which has globals parameters was for twas_cv, that some additional argument will need to be included (such as sample_partition,  args, cv_args) as the facilitating function is only asking for fold number but it would need additional information from the environment. 
`compute_method_predictions <- function(j){...}`

R/mrmash_wrapper.R
gaow: sounds good then!

R/mrmash_wrapper.R
gaow: It looks reasonable to me now ... i cannot think of counter examples that this would fail.

PR #233, opened 2024-08-30
update filter_X

Chunmingl:


R/multivariate_pipeline.R
```diff
@@ -242,8 +242,10 @@ multivariate_analysis_pipeline <- function(
 
   # filter X based on Y subjects
   if (!is.null(imiss_cutoff) || !is.null(maf_cutoff)) {
-    X <- filter_X_with_Y(X, Y, imiss_cutoff, maf_cutoff, var_thresh = xvar_cutoff, X_variance = X_variance)
-    maf <- maf[colnames(X)]
+    variants_prefilter <- colnames(X)
```
gaow, line 245: This is not necessary since MAF is named

R/univariate_pipeline.R
```diff
@@ -93,7 +93,7 @@ univariate_analysis_pipeline <- function(
   # Filter X based on missingness, MAF, and variance
   if (!is.null(imiss_cutoff) || !is.null(maf_cutoff)) {
     # FIXME: I think we should use filter_X directly for univariate without considering Y? 
-    X_filtered <- filter_X(X, imiss_cutoff, maf_cutoff, var_thresh = xvar_cutoff, X_variance = X_variance)
+    X_filtered <- filter_X(X, imiss_cutoff, maf_cutoff, var_thresh = xvar_cutoff, maf = maf, X_variance = X_variance)
```
gaow, line 96: We need it here only because x_scalar is unnamed.

R/multivariate_pipeline.R
Chunmingl: ok

R/univariate_pipeline.R
Chunmingl: do we need any additional changes at this ?

PR #245, opened 2024-09-16
update ctwas workflow and data loader

Chunmingl:


R/ctwas_wrapper.R
```diff
@@ -14,47 +14,132 @@ get_ctwas_meta_data <- function(ld_meta_data_file, regions_table) {
...
+          #R_wgt = post_qc_twas_data[[gene]]$LD[context_variants, context_variants, drop = FALSE], # ld_list$combined_LD_matrix[context_variants, context_variants],
+          molecular_id = gene,
+          weight_name = paste0(data_type, "_", context),
+          type = data_type,
+          context = context,
+          n_wgt = length(context_variants)
+        )
+      }
+    }
+    # weights_list <- compute_weight_LD_from_ref(weights_list,
```
gaow, line 120: As i said we don't know why we want to compute weight LD at this point. The input you prepare should not have any LD information included. It will be dealt with in cTWAS codes anyways, right?

PR #248, opened 2024-09-19
merge twas, ctwas, and twas_mr, clean up

Chunmingl:
work in progress


R/ctwas_wrapper.R
```diff
@@ -23,38 +32,97 @@ get_ctwas_meta_data <- function(ld_meta_data_file, regions_table) {
...
+    genes <- names(post_qc_twas_data)
+    # Compile weights_list
+    weights_list <- list()
+    for (gene in genes) {
+      contexts <- names(post_qc_twas_data[[gene]][["weights_qced"]])
+      gene_data <- twas_weights_data[[gene]][["export_twas_weights_db"]]
+      for (context in contexts) {
+        data_type <- post_qc_twas_data[[gene]][["data_type"]][[context]]
+        postqc_scaled_weight <- post_qc_twas_data[[gene]][["weights_qced"]][[context]][["scaled_weights"]]
+        # adjust susie weights for selected number of weight variants
```
gaow, line 51: This step must have already been done in harmonize TWAS weights before you do TWAS? The scaled weights should already have been computed based on the correct SuSiE weights.

R/twas.R
```diff
@@ -240,18 +239,16 @@ harmonize_twas <- function(twas_weights_data, ld_meta_file_path, gwas_meta_file,
   }
 
   # Step 1: load TWAS weights data
-  loader <- twas_data_loader$new(twas_weights_data)
-  twas_weights_data <- loader$get_data()
-  genes <- names(twas_weights_data)
-  chrom <- as.integer(twas_weights_data[[1]]$chrom)
+  genes <- twas_weights_data$gene
+  chrom <- as.integer(readr::parse_number(gsub(":.*$", "", twas_weights_data[["export_twas_weights_db"]][[1]][["variant_names"]][1])))
```
gaow, line 243: use Roxygen2 `importFrom readr parse_number`. Dont use `::` here.

R/twas.R
```diff
@@ -353,46 +351,70 @@ harmonize_twas <- function(twas_weights_data, ld_meta_file_path, gwas_meta_file,
...
         )
+        # summarize twas results by methods for a context within a gene
         context_table <- data.frame(
           gwas_study = study, method = sub("_[^_]+$", "", names(twas_rs)), twas_z = find_data(twas_rs, c(2, "z")),
           twas_pval = find_data(twas_rs, c(2, "pval"))
         )
         return(context_table)
       }))
+      # mr analysis - for a single context
+      mr_formatted_input <- mr_format(twas_weights_results[[gene]]$susie_results, context, twas_data_qced[[gene]][["gwas_qced"]][[study]],
```
gaow, line 388: For MR, I think the idea is to only do it for those that are TWAS significant.

gaow: You can filter for that with some user specified TWAS p-value cutoff and only perform MR beyond that cutoff.

R/ctwas_wrapper.R
Chunmingl: yes, you're right, this is redundant, and should be removed on the adjust_susie_weights step

R/twas.R
Chunmingl: ok

R/twas.R
Chunmingl: and set default p-value cutoff for mr analysis be 0.05?

gaow:
0.05 is acceptable although a bit too loose. But it's ok for this purpose
On the go. Sorry for short text.

PR #250, opened 2024-09-20
replace model selections argument

Chunmingl:


R/twas.R
```diff
@@ -462,14 +463,17 @@ twas_pipeline <- function(twas_weights_data,
         twas_mr_rs$gwas_study <- study
         mr_table <- rbind(mr_table, twas_mr_rs)
       }
+      model_selections[[molecular_id]][[context]] <- twas_weights_data[[molecular_id]][["export_twas_weights_db"]][[context]][["selected_model"]]
```
gaow, line 466: I dont understand why you need to use this variable. Why don't you just filter here on `twas_data_qced` to only keep the model selected so you remove the need of this paramter algother? That was my original suggestion.

Chunmingl: I agree, this is much cleaner approach!
This is now addressed, please have a look,

PR #252, opened 2024-09-23
revise load_multitrait_R_sumstats

AnqiWang2021:


R/mash_wrapper.R
```diff
@@ -470,16 +470,59 @@ load_multitrait_R_sumstat <- function(
   }
 
   merge_matrices <- function(matrix_list, value_column, id_column = "variants",
-                             remove_any_missing = FALSE) {
-    # Convert matrices to data frames
-    df_list <- lapply(seq_along(matrix_list), function(i) {
+                             remove_any_missing = FALSE, ld_ref = ld_ref, allele_flip = TRUE) {
```
gaow, line 473: `ld_ref` cannot default to a variable -- this can be considered a programming error. If this is required input, you can move it up to before `id_column` and make it required without default value.

R/mash_wrapper.R
```diff
@@ -470,16 +470,59 @@ load_multitrait_R_sumstat <- function(
...
+      # Step 1: Convert matrix to data frame and extract relevant columns
       df <- as.data.frame(matrix_list[[i]])
       df2 <- df[, c(id_column, value_column)]
-      # Rename columns to avoid duplication
-      colnames(df2) <- c(id_column, paste0(value_column, "_", i))
-      return(df2)
-    })
-
+      if (allele_flip) {
+       # Step 2: Split 'variants' to extract chromosomal info
```
gaow, line 479: can you define this as a function called `parse_variant_id()` similar to our `parse_region` and put it right under: https://github.com/cumc/pecotmr/blob/main/R/misc.R#L218 ?

R/mash_wrapper.R
```diff
@@ -470,16 +470,59 @@ load_multitrait_R_sumstat <- function(
...
+       # Step 4: Merge with LD reference and filter
+       variants_ld_block_match <- merge(cohort_df, ld_ref, by = "chrom", allow.cartesian = TRUE)%>%
+                                 filter(pos > start & pos < end) %>%
+                                 select(-path)
+          
+       # Function to process each group
+       process_group <- function(data) {  
+          # Construct file path
+          bim_file_path <- unique(data$bim_path)
+          ld_bim_file_path <- paste0("/mnt/vast/hpc/csg/data_public/20240120_ADSP_LD_matrix/",bim_file_path, sep="")
```
gaow, line 500: you can't use hard-coded path here

R/mash_wrapper.R
```diff
@@ -470,16 +470,59 @@ load_multitrait_R_sumstat <- function(
...
+    df_list <- lapply(seq_along(matrix_list), function(i){
+      # Step 1: Convert matrix to data frame and extract relevant columns
       df <- as.data.frame(matrix_list[[i]])
       df2 <- df[, c(id_column, value_column)]
-      # Rename columns to avoid duplication
-      colnames(df2) <- c(id_column, paste0(value_column, "_", i))
-      return(df2)
-    })
-
+      if (allele_flip) {
```
gaow, line 478: can we call this argument `allele_qc = TRUE` not `allele_flip = TRUE`?

R/mash_wrapper.R
```diff
@@ -416,70 +416,99 @@ merge_susie_cs <- function(susie_fit, coverage = "cs_coverage_0.95", complementa
...
-    bhat <- as.data.table(cbind(sumstats_db$variant_names, sumstats_db$sumstats$betahat))
-    sbhat <- as.data.table(cbind(sumstats_db$variant_names, sumstats_db$sumstats$sebetahat))
-    setnames(bhat, c("variants", "bhat"))
-    setnames(sbhat, c("variants", "sbhat"))
-    bhat[, bhat := as.numeric(bhat)]
-    sbhat[, sbhat := as.numeric(sbhat)]
-    list(bhat = bhat$bhat, sbhat = sbhat$sbhat, variants = bhat$variants)
-  }
-
+load_multitrait_R_sumstat <- function(susie_fit,sumstats_db, coverage = NULL, top_loci = FALSE, filter_file = NULL, remove_any_missing = TRUE, max_rows_selected = 300, nan_remove=FALSE, exclude_condition, ld_meta_file) {
```
gaow, line 419: All required argument without default parameters should be moved to before all the arguments with default paramteters. Also as I suggested ld_meta_file should be default to NULL which means no allele QC needed. So the interface should be:

```
load_multitrait_R_sumstat <- function(susie_fit,sumstats_db, exclude_condition=NULL, ld_meta_file = NULL, coverage = NULL, top_loci = FALSE, filter_file = NULL, remove_any_missing = TRUE, max_rows_selected = 300, nan_remove=FALSE)
```

and later you will have 

```r
if (!is.null(ld_meta_file)) {}
if (!is.null(exclude_condition)) {}
```

R/mash_wrapper.R
```diff
@@ -492,70 +521,61 @@ load_multitrait_R_sumstat <- function(
...
+  out$bhat <- out$bhat[var_idx,,drop=FALSE]
+  out$sbhat <- out$sbhat[var_idx,,drop=FALSE]
+  
+   rownames(out$bhat) <- rownames(out$sbhat) <- variants
+   colnames(out$bhat)[2:ncol(out$bhat)] <- colnames(out$sbhat)[2:ncol(out$bhat)] <- trait_names
+   out$bhat <- out$bhat[,-which(names(out$bhat)=="variants"), drop = FALSE]
+   out$sbhat <- out$sbhat[,-which(names(out$sbhat)=="variants"), drop = FALSE]
+   out$region = names(susie_fit)
+                    
+   if (length(exclude_condition) > 0) {
```
gaow, line 567: as I suggested earlier, you can make `exclude_condition = NULL` by default and here use `is.null` to check it.

PR #262, opened 2024-10-01
temporary remove bayesX methods

Chunmingl:


gaow:
@Chunmingl tbh i think Peter might beat us in terms of fixing the issue so i'll keep them here for now without merging this PR but to close it. When we need to merge i'll reopen and merge.

PR #281, opened 2024-10-13
add function z_to_beta_se

RuiDongDR:
function `z_to_beta_se` is designed to impute `beta` and `se` when there is only z-scores in the summary statistics (sample size and MAF required too).


gaow:
@RuiDongDR nice documentation! I've messaged you some comments on slack. Let's address from there to improve the PR. Thanks!

PR #287, opened 2024-10-21
minor fix on syntax error

al4225:


gaow:
Thanks @al4225 i made some minor changes yesterday without testing. Good you catch that!

PR #291, opened 2024-10-24
rsq and pval name option and minor fixes

Chunmingl:


R/twas.R
```diff
@@ -252,6 +252,8 @@ twas_pipeline <- function(twas_weights_data,
                           p_val_cutoff = 0.05,
                           quantile_twas = FALSE, 
                           select_best_model = TRUE,                          
+                          rsq_option = "rsq",
```
gaow, line 255: here you should make it also `c('rsq','adj_rsq')`?

R/twas.R
```diff
@@ -158,6 +158,7 @@ harmonize_twas <- function(twas_weights_data, ld_meta_file_path, gwas_meta_file)
       for (study in names(gwas_files)) {
         gwas_file <- gwas_files[study]
         gwas_sumstats <- as.data.frame(tabix_region(gwas_file, query_region)) # extension for yml file for column name mapping
+        if(nrow(gwas_sumstats)==0) return (list(twas_data_qced = NULL, snp_info = NULL))
```
gaow, line 161: This is nice. But I wonder if you can always add a message to it, like:

```
warning(paste("No GWAS summary statistics are found in the region of interest", query_region))
```

(you may want to format the region better)

Every time you return something like NULL it is best to include a message . Otherwise we may wonder what was causing the NULL

R/twas.R
```diff
@@ -370,6 +371,7 @@ twas_pipeline <- function(twas_weights_data,
     twas_weights_data[[weight_db]][["molecular_id"]] <- weight_db
     twas_data_qced_result <- harmonize_twas(twas_weights_data[[weight_db]], ld_meta_file_path, gwas_meta_file)
     twas_data_qced <- twas_data_qced_result$twas_data_qced
+    if(is.null(twas_data_qced)) return(NULL)
```
gaow, line 374: Same issue here. you need to comment with a warning that "No imputable gene-context pairs are left"; then return NULL

PR #314, opened 2024-12-16
MASH_Handle different structures

rl3328:
1. main changes in load_multitrait_R_sumstat(): a. add a  filter to only restrict to conditions in finemapped tsv file, not go over the whole rds file. b. felxibly handle different structure of xqtl and metabrain, go deeper detect $variant_names and $sumstats layer and convert bhat, sbhat into z score.(also made variants format consistent for xqtl and metabrain, which will otherwise cause error when merging in a flexible way)
2. changed all bhat, sbhat as z, which cause changes in large chunck of code. This might affect other functions(e.g. functions use filter_invalid_summary_stat() ).
3. I have no idea how to solve the erroneous checks...
4. All my modified functions were flexible to handle both "z" and "bhat/sbhat" structure. filter_invalid_summary_stat() was called in multigene_udr.R seperately without load_multitrait_R_sumstat(), I modified the commands accordingly. mash_rand_null_sample() is also improved, although til now, no other .R call it seperately.
In summary, I think the code should be good for handling these two structures.


R/mash_wrapper.R
```diff
@@ -518,59 +594,84 @@ load_multitrait_R_sumstat <- function(susie_fit, sumstats_db, coverage = NULL, t
   }
 
   merge_matrices <- function(matrix_list, value_column, ld_meta_file, id_column = "variants",
-                             remove_any_missing = FALSE) {
+                               remove_any_missing = FALSE) {
+    # Input validation
+    if (!is.list(matrix_list) || length(matrix_list) == 0) {
+      stop("matrix_list must be a non-empty list")
+    }
+    if (!is.character(value_column) || length(value_column) != 1) {
```
rl3328, line 602: just input validation, can be ignored

R/multigene_udr.R
```diff
@@ -69,7 +69,7 @@ multigene_udr <- function(combined_susie_list, coverage, independent_variant_lis
       mash_input,
       bhat = paste0(cond, ".b"),
       sbhat = paste0(cond, ".s"),
-      z = TRUE,
+      btoz = TRUE,
```
rl3328, line 72: filter_invalid_summary_stat() is also called here, so I changed the commands here accordingly

PR #315, opened 2024-12-17
fix unit test failure for twas.R

Chunmingl:


Chunmingl:
`Test null variable_name_obj` is commented out because `variant_name_obj` cannot be NULL

PR #330, opened 2025-02-04
update rss_finemapping part

rl3328:
1. Added a parameter no_comment to decide whether to to consider all content after '#' as comment in the column_mapping file(need corresponding modifications in rss_analysis.ipynb)
2. Fixed the `unused argument median_abs_corr = median_abs_corr`  error in some phenotypes
3. Tested and the output looked fine.

### DIFF
--- a/R/file_utils.R
+++ b/R/file_utils.R
@@ -762,7 +762,7 @@ load_twas_weights <- function(weight_db_files, conditions = NULL,
 #' @param region The region where tabix use to subset the input dataset.
 #' @param target User-specified gene/phenotype name used to further subset the phenotype data.
 #' @param target_column_index Filter this specific column for the target.
-#' @param no_comment whether to consider all content after '#' as comment in the column_mapping file.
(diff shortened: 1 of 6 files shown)


gaow:
@rl3328 thanks, but two issues:

1. I want to create a parameter for comment string as `comment_string = "#"` by default. Then when `comment_string=NULL` it will skip comment. In the SoS interface we just set it `paramter: comment_string = "NULL"`
2. Please revert the change on the median corr argument. If you use `gaow/susieR` it has the median correlation and that version will be used as future default SuSiE. Let's keep it that way, don't change it.

rl3328:
But there's one error only in linux:
```
Error: Failure in `/tmp/RtmptMVxGO/R_LIBS30386024cdd5/pecotmr/pecotmr-tests/testthat.Rout.fail`
 parallel processing ──
result_parallel$prediction$glmnet_predicted not equal to result_single$prediction$glmnet_predicted.
30/30 mismatches (average diff: 0.178)
```

PR #333, opened 2025-02-08
Fix a typo in load_rss_data interms of 'comment_string'

rl3328:
1. I fixed a typo to make comment-string considered as a char. This typo didn't affect my analyses, but for others which use comment.char, it might cause some issues.
I only changed `if (is.null(comment_string)) {` to `if (comment_string == "NULL")) {`, other changs show because that I commit in a new branch.


R/file_utils.R
```diff
@@ -773,7 +773,7 @@ load_twas_weights <- function(weight_db_files, conditions = NULL,
 load_rss_data <- function(sumstat_path, column_file_path, subset = TRUE, n_sample = 0, n_case = 0, n_control = 0, target = "",
                           region = "", target_column_index = "", comment_string = "#") {
   # Read and preprocess column mapping
-  if (is.null(comment_string)) {
```
gaow, line 776: @rl3328 i don't this this change is valid. We can talk

PR #343, opened 2025-02-20
make some dependencies optional

danielnachun:
This is a very large PR to make many of our dependencies optional.   This will simplify installation of the package, especially from source.  In local testing, all the unit tests that passed before still pass - we have accumulated some test failures unrelated to this PR which I will not be fixing here.


.github/recipe/recipe.yaml
```diff
@@ -51,15 +47,13 @@ requirements:
     - r-qgg
     - r-quadprog
     - r-quantreg
-    - r-r.utils
```
gaow, line 54: Im not sure where this is used but this sounds like a very basic package?

DESCRIPTION
```diff
@@ -13,48 +13,48 @@ Authors@R: c(person("Gao Wang",role = c("cre","aut"),
...
-    mr.mash.alpha,
-    mvsusieR,
-    pgenlibr,
     purrr,
-    qgg,
-    quantreg,
-    qvalue,
     readr,
     rlang,
-    snpStats,
```
gaow, line 47: if we have bigstatsr not sure where we need snpStats --- did we use snpStats in the genotype data loader from PLINK format? if so, we still need that. That is currently the only tested way to load genotype from

.github/recipe/recipe.yaml
danielnachun: It has no imports any more, so whatever we did use it for is now gone.

DESCRIPTION
danielnachun: Yes, it's for loading PLINK formats, and VCFs which I will add support for later.

R/file_utils.R
```diff
@@ -120,36 +126,39 @@ NoSNPsError <- function(message) {
 #' @param keep_indel Whether to keep indel SNPs.
 #' @return A vector of SNP IDs in the specified region.
 #'
-#' @importFrom data.table fread
+#' @importFrom vroom vroom
 #' @importFrom magrittr %>%
-#' @importFrom snpStats read.plink
 #' @export
 load_genotype_region <- function(genotype, region = NULL, keep_indel = TRUE, keep_variants_path = NULL) {
+  # Make sure snpStats is installed
```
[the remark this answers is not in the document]
gaow: Actually we may need to add to our pixi-setup script all the dependencies we now removed?

R/mash_wrapper.R
```diff
@@ -923,22 +928,23 @@ mash_pipeline <- function(mash_input, alpha, residual_correlation = NULL, uncons
...
     Shat = mash_input$strong.s, V = vhat,
     alpha, zero_Bhat_Shat_reset = 1000
   )
   # Canonical matrices
-  U.can <- cov_canonical(mash_data)
+  U.can <- mashr::cov_canonical(mash_data)
 
   # Penalty strength
   lambda <- ncol(mash_input$strong.z)
+  # NOTE: Please change this to use flashier + ED instead of UDR
```
gaow, line 947: @danielnachun this is very considerate -- could you use FIXME instead of NOTE? I typically do that then at some point search for all FIXME string in the code repo to see what needs extra work

R/regularized_regression.R
```diff
@@ -436,15 +439,14 @@ lasso_weights <- function(X, y) glmnet_weights(X, y, 1)
...
-#' @importFrom mr.ash.alpha mr.ash
 #' @importFrom stats predict
 #' @export
 mrash_weights <- function(X, y, init_prior_sd = TRUE, ...) {
   args_list <- list(...)
   if (!"beta.init" %in% names(args_list)) {
     args_list$beta.init <- lasso_weights(X, y)
   }
-  fit.mr.ash <- do.call("mr.ash", c(list(X = X, y = y, sa2 = if (init_prior_sd) init_prior_sd(X, y)^2 else NULL), args_list))
+  fit.mr.ash <- do.call("mr.ash.alpha::mr.ash", c(list(X = X, y = y, sa2 = if (init_prior_sd) init_prior_sd(X, y)^2 else NULL), args_list))
```
gaow, line 449: we need to check for existance of mr.ash.alpha at the start of this function

R/susie_ash.R
```diff
@@ -195,7 +195,6 @@
 #'
 #' @importFrom stats var
 #' @importFrom utils modifyList
-#' @importFrom mr.ash.alpha mr.ash coef.mr.ash
```
gaow, line 198: @danielnachun let's remove this file from pecotmr. We have a separate package for it now and eventually this is going into susieR

R/file_utils.R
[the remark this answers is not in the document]
danielnachun: It's a still a dependency of the `conda` package!  Just not in the NAMESPACE file.  So it will have no effect on our current setup - it will only matter when installing from source without a pixi environment.

R/regularized_regression.R
danielnachun: Thanks, I missed this one!

R/susie_ash.R
danielnachun: I'll remove this in my next commit.

PR #350, opened 2025-02-23
Initial draft of colocboost pipline

xueweic:
Create colocboost_pipline.R
Initial draft of colocboost_pipline: including
1. data_loader from multiple dataset
2. data_qc by PIP, maf, sumstatQC
3. colocboost_analysis code

Please don't merge it now. I have several questions wanting to discuss with you first.


xueweic:
Added the sQTL filtering before run ColocBoost: i) keep the clusters including PR. ii) remove the clusters including IN.

xueweic:
Update function names and input names

R/colocboost_pipline.R
```diff
@@ -0,0 +1,667 @@
...
+        return(y)
+    })
+    names(res) <- names(res_Y)
+    return(res)
+}
+
+#' Remove variants with maf < maf_cutoff
+#' @return A list of residual X and maf
+#' @export
+filter_resX_maf <- function(res_X, maf, maf_cutoff = 0.0005){
```
gaow, line 318: We have this function already: https://github.com/StatFunGen/pecotmr/blob/main/R/misc.R#L163 when MAF is provided it will filter by the provided MAF instead of recomputing it.

R/colocboost_pipline.R
```diff
@@ -0,0 +1,667 @@
...
+        res_X <- res_X[, -pos, drop = FALSE] %>% .[, order(colnames(.)), drop = FALSE]
+        maf <- maf[-pos]
+    }
+    return(list(res_X = res_X, maf = maf))
+}
+
+#' Initial PIP check for individual level data
+#' @return A list of residual Y
+#' @export
+filter_resY_pip <- function(res_X, res_Y, pip_cutoff_to_skip = 0, context = NULL){
```
gaow, line 330: We also have something similar here: https://github.com/StatFunGen/pecotmr/blob/main/R/multivariate_pipeline.R#L99 if you think it is cleaner to keep using your customized version I would suggest hiding this function into the pipeline function itself (just like this `skip_conditions` function here) not making it an explicit function that gets exported.

R/colocboost_pipline.R
```diff
@@ -0,0 +1,667 @@
...
+                           
+#' sumstat_data QC, imputation using raiss
+#' @return A list of sumstat_data after initial checking and QC:
+#' \itemize{
+#'   \item sumstats: A list of summary statistics and ready to do analysis.
+#'   \item LD_mat: A list of LD matrix and ready to do analysis.
+#'   \item LD_match: A vector of strings to indicating sumstats and LD matching (save space since multiple sumstats may link to the same LD matrix).
+#' }
+#' @export     
+sumstat_qc <- function(sumstat_data,
```
gaow, line 476: we already also have this function: https://github.com/StatFunGen/pecotmr/blob/main/R/sumstats_qc.R#L145 it is very confusing to create these other similar, redundant functions. Please consolidate.

R/colocboost_pipline.R
```diff
@@ -0,0 +1,667 @@
...
+            message(paste("Skipping follow-up analysis for all individual traits", paste(names(Y)[null_Y], collapse = ";"), "."))
+            X <- NULL
+            Y <- NULL
+        }
+        if (!is.null(Y)){
+            Y <- lapply(1:length(Y), function(i){
+                y <- Y[[i]]
+                events <- colnames(y)
+                condition <- names(Y)[i]
+                if (all(grepl("sQTL", events))){
```
gaow, line 586: what's special about sQTL? To me, we filter our events for two criteria: not enough sample size (that you can deduce from the data) or failing some PIP cutoff in initial screen. This special treatment should not be needed for sQTL. If you want to exclude some events based on their names then you can add the third criteria but has to be generic

gaow: For example you can pass the regular expression (`regex`) that filters the patterns you don't like. I guess `sQTL_*_IN*` for example but has to be some regular expression that actually works

NAMESPACE
```diff
@@ -41,6 +45,7 @@ export(harmonize_twas)
 export(lasso_weights)
 export(lbf_to_alpha)
 export(load_LD_matrix)
+export(load_colocboost_regional_data)
```
gaow, line 48: this, and the colocboost pipeine function are the only two that should be exported. Perhaps not others

R/colocboost_pipline.R
```diff
@@ -0,0 +1,667 @@
...
+#'   \item Y_coordinates: Phenotype coordinates if a region is specified.
+#' }
+#' sumstat_data contains the following components if exist
+#' \itemize{
+#'   \item sumstats: A list of summary statistics for the matched LD_info, each sublist contains sumstats, n, var_y from \code{load_rss_data}.
+#'   \item LD_info: A list of LD information, each sublist contains combined_LD_variants, combined_LD_matrix, ref_panel  \code{load_LD_matrix}.
+#' }
+#'
+#' @export
+load_colocboost_regional_data <-  function(region, # a string of chr:start-end for phenotype region
```
gaow, line 70: okay we decide to finalize on `load_multitask_regional_data`

R/colocboost_pipline.R
xueweic: I see. I am looking in this univariate analysis pipline. For multi-task QC, we may have the same LD with different sumstats, for example, we only need to look LD matrix one time for 7 AD GWAS sumstat, but after QC and taking the overlapped variants. LD may or may not as the same. Could we rename it as `sumstat_qc_multitask` and put in `sumstat_qc.R` file?

https://github.com/StatFunGen/pecotmr/blob/0a9200f889ea325ea375291e40105c35585fa77d/R/univariate_pipeline.R#L213-L225

R/colocboost_pipline.R
xueweic: changed by adding
sqtl_pattern = "clu_(\\d+_[+-?]):"
 include_context = "clu_(\\d+_[+-?]):PR:"
 exclude_context = "clu_(\\d+_[+-?]):IN:"

xueweic:
@gaow changed the following

1. only export two functions `load_multitask_region_data` and `colocboost_analysis_pipline`. All other functions are internal or noRd.
2. adding three regex to colocboost_analysis_pipline to process sQTL.
3. remove filter_resX_maf, instead of using filter_X in misc.R

R/colocboost_pipline.R
```diff
@@ -0,0 +1,677 @@
...
+    res_cb <- colocboost(X = X, Y = Y, sumstat = sumstats, LD = LD_mat,
+                         dict_YX = dict_YX, dict_sumstatLD = dict_sumstatLD, 
+                         traits_names = traits, target_idx = target_idx, ...)
+                                                     
+    return(res_cb)                        
+
+}
+
+
+#' The following functions will go to colocboost_pipline.R
```
gaow, line 389: remove this line too

xueweic:
@gaow removed two lines from colocboost pipline

PR #355, opened 2025-02-27
Update colocboost_pipline.R

xueweic:
Fixed several bugs with the current pipeline. This analysis code works most cases. Please review and merge it.


xueweic:
adding computing time saving

xueweic:
fixed bug for removing all events based on events rules

PR #358, opened 2025-03-14
minor fix for namespace not loading within "do.call"

Chunmingl:


R/regularized_regression.R
```diff
@@ -450,7 +450,7 @@ mrash_weights <- function(X, y, init_prior_sd = TRUE, ...) {
   if (!"beta.init" %in% names(args_list)) {
     args_list$beta.init <- lasso_weights(X, y)
   }
-  fit.mr.ash <- do.call("mr.ash.alpha::mr.ash", c(list(X = X, y = y, sa2 = if (init_prior_sd) init_prior_sd(X, y)^2 else NULL), args_list))
+  fit.mr.ash <- mr.ash.alpha::mr.ash(X = X, y = y, sa2 = if (init_prior_sd) init_prior_sd(X, y)^2 else NULL, args_list)
```
danielnachun, line 453: If you can change:
```
args_list <- list(...)
```
to
```
args_list <- list2(...)
```
and then make this change

```suggestion
  fit.mr.ash <- mr.ash.alpha::mr.ash(X = X, y = y, sa2 = if (init_prior_sd) init_prior_sd(X, y)^2 else NULL, !!!args_list)
```

It should work for the arguments

I think you'll also then need:
```
#' @importFrom !!! rlang
#' @importFrom list2 rlang
```

Chunmingl: I tried with the updated function below 

```
mrash_weights <- function(X, y, init_prior_sd = TRUE, ...) {
  # Make sure mr.ash is installed
  if (! requireNamespace("mr.ash.alpha", quietly = TRUE)) {
    stop("To use this function, please install mr.ash: https://github.com/stephenslab/mr.ash.alpha")
  }
  args_list <- list2(...)
  if (!"beta.init" %in% names(args_list)) {
    args_list$beta.init <- lasso_weights(X, y)
  }
  fit.mr.ash <- mr.ash.alpha::mr.ash(X = X, y = y, 
              sa2 = if (init_prior_sd) init_prior_sd(X, y)^2 else NULL, !!!args_list)
  predict(fit.mr.ash, type = "coefficients")[-1]
}
```
and I got error message of 
```
Error in `map()`:
i In index: 5.
Caused by error in `!args_list`:
! invalid argument type
Traceback:
```
initially I thought it was because the step that adds additional list of $beta.init but got the same error by unlist with !!! and list2 again after adding beta.init 
`args_list$beta.init <- lasso_weights(X, y)`

Chunmingl: I tried as below and it seemed to work, please advise 
```
mrash_weights <- function(X, y, init_prior_sd = TRUE, ...) {
  # Make sure mr.ash is installed
  if (! requireNamespace("mr.ash.alpha", quietly = TRUE)) {
    stop("To use this function, please install mr.ash: https://github.com/stephenslab/mr.ash.alpha")
  }
  args_list <- list2(...)
  if (!"beta.init" %in% names(args_list)) {
    args_list$beta.init <- lasso_weights(X, y)
  }
  fit.mr.ash <- do.call(getFromNamespace("mr.ash", "mr.ash.alpha"), 
                c(list(X = X, y = y, sa2 = if (init_prior_sd) init_prior_sd(X, y)^2 else NULL), args_list))
  predict(fit.mr.ash, type = "coefficients")[-1]
}
```

danielnachun: Okay that work around is actually fine, and you can switch `list2` back to `list`.

PR #364, opened 2025-03-23
minor fix bugs

xueweic:
1. In LD.R, exchange `arrange` and `check order`. This is the same logic in the previous code, otherwise, we need to input an ordered `ld_meta_file`. - Met an error in all genes, since our ld_meta_file is not ordered by chr and pos.
2. In raiss.R, if sumstat does not contain known varaints (raiss_single_matrix.R), it return NULL. we can not check `nrow`. - Met an error in chr1_ENSG00000143418 and chr1_ENSG00000143437 in Kunkle sumstat. Kunkle does not have variants in the first LD block, then it reported an error `Error in if (nrow(combined_result) == 0) { : argument is of length zero`.


gaow:
>  Met an error in all genes, since our ld_meta_file is not ordered by chr and pos.

okay i caught this when i was working on the code yesterday thinking that if we "arrange" it already then the consecutive check would not make sense. Maybe this function is not that useful after all anyways and it is cleaner we remove that check.

PR #367, opened 2025-03-25
map data type function for sQTL contexts

Chunmingl:


R/file_utils.R
```diff
@@ -980,3 +980,23 @@ batch_load_twas_weights <- function(twas_weights_results, meta_data_df, max_memo
...
+
+#' Map data type for twas weight contexts based on provided data type table. 
+#' @param data_type_table A data frame contains three columns: type, context, subgroup 
+#' @param context_names A vector of contexts 
+#' @return A vector of mapped data type based on context names
+#' @export
+map_data_type <- function(data_type_table, context_names){
+        data_type <- lapply(context_names, function(context){
+             xqtl_type_table$type[apply(xqtl_type_table, 1, function(x) {
+                 if (! grepl("sQTL",context)) {
```
gaow, line 993: @Chunmingl am not entirely sure what we are trying to do here but anything hard-coded is not good. If you want to map data, we should take proper regular expression (search for " `regex` in R" ) then we provide some external files called `qtl_type_mapping` that look like this:

```
sQTL, "clu_(\\d+_[+-.]):"
```

which is the pattern for sQTL event names.

gaow: just as an exmaple. Then you load this external file and use this to help you identify the correct data types

Chunmingl: got it, I will update the function using `regex`

PR #374, opened 2025-04-10
archive: first use rename, if failed, use file.copy()+file.remove(), …

al4225:
archive: first use rename, if failed, use file.copy()+file.remove(), also message warnings.

In cloud environments, the source directory and target directory are often located on different file systems or partitions. When attempting to move files across different file systems using the `file.rename()` function, the operating system returns an "Invalid cross-device link" error. This occurs because the rename system call can only operate within the same file system, as it's designed to simply update directory entries rather than physically move data between storage devices.


gaow:
I see! okay that makes sense. Then indeed you have to copy and delete. That seems the only way to work it with S3 bucket.

PR #386, opened 2025-04-30
Update coloc related functions 

rfeng2023:
This commit includes:
1.	Match variants in `allele_qc`
2.	Add sample LD mode in `load_and_extract_ld_matrix`, since the LD reference panel is no longer appropriate for two xQTL modalities. This is using original genotype data instead of residual genotype data.
3.	Add a correlation function `get_cormat` from ColocBoost, for above in sample LD calculation
4.	Minor fix to get_nested_element to allow '' as input and skip that layer — useful for multi-context/gene structures with different layer hierarchies


R/allele_qc.R
```diff
@@ -130,6 +130,7 @@ allele_qc <- function(target_data, ref_variants, col_to_flip = NULL,
     match_result[strand_flipped_indices, "A2.target"] <- strand_flip(match_result[strand_flipped_indices, "A2.target"])
   }
 
+
```
gaow, line 133: What should we do with this FIXME?

gaow:
Thanks Ru I think the code can be made more general in merging fsusie and SuSiE input, and when LD matrices are Indeed loaded we should be able to do a better size check using purity in the filter LBF part as we discussed yesterday. But it's fine for the time being. I might work with Hao for another round of improvement next week

R/allele_qc.R
rfeng2023: Here was just a space character accidentally added during a previous modification—nothing substantial was changed (I went through several attempts to fix it, but eventually realized that simply keeping the original ID was the easiest solution.)

gaow:
Thanks @rfeng2023 so this patch is good to merge based on your standards? Just wonder if there are still any known issues.

rfeng2023:
> Thanks @rfeng2023 so this patch is good to merge based on your standards? Just wonder if there are still any known issues.

yes please @gaow

PR #503, opened 2026-05-29
simple fix - removing `colocboost::`

xueweic:


danielnachun:
I think we also need to update the `DESCRIPTION` file to put `colocboost` in `Imports`, don't we?

PR #504, opened 2026-05-30
Unify compact top-loci export and SuSiE-Inf as the option for univaraite pipeline

xueweic:
## Unify SuSiE top-loci output: one helper, one table

**Major**
- Replace `build_top_loci_long` / `_wide` / `_export` trio with single `build_top_loci()` returning one 22-column table.
- `postprocess_finemapping_fits()` runs one method for-loop and rbinds per-method rows; `format_finemapping_output()` exposes only `top_loci`.

**Minor**
- `build_top_loci()` keeps the old long-builder input shape; `data_x` / `data_y` / `other_quantities` stay default-NULL.
- `other_quantities` reserves `region` and `condition_id` (alongside `dropped_samples`).
- `FineMappingResult` S4 slot keeps legacy shape via back-compat alias inside `susie_wrapper.R`; `AllClasses.R` / `AllMethods.R` / `vcf_writer.R` unchanged.
- Adds hard gating tests: 22-col order + dtypes, `<method>_<idx>` CS strings, `<method>_0` PIP-only, per-method independent CS numbering, `cs_95_purity = 0` rule, overlapping CS within/across methods, single `top_loci` field, removed-trio absence.

**Final `top_loci` (22 cols)**
| Group | Columns |
|---|---|
| variant | `#chr`, `start`, `end`, `a1`, `a2`, `variant` |
| phenotype | `gene`, `event` |
| marginal | `n`, `maf`, `beta`, `se` |
| fine-mapping | `pip`, `posterior_effect_mean`, `posterior_effect_se`, `cs_95`, `cs_70`, `cs_50`, `cs_95_purity` |
| run-level | `method`, `grange_start`, `grange_end` |

**Conventions**
- `cs_95` / `cs_70` / `cs_50` are character `"<method>_<idx>"` (e.g. `"susie_1"`, `"susie_inf_2"`); PIP-only retained variants use `"<method>_0"`. Each method numbers CS independently from 1.
- `cs_95_purity` = 0.95-coverage purity for the row's `(method, cs_95)`; `"<method>_0"` rows carry `0`.
- Row uniqueness: `(variant, gene, method, cs_membership)` — overlapping CS within a method and the same variant across methods both produce separate rows.

## SuSiE-Inf as the option for univaraite pipeline

Univariate fine-mapping can now bypass SuSiE-inf initialization when ordinary SuSiE fitting is explicitly requested. The SuSiE-inf path remains the default for compatibility, and TWAS-weight generation continues to require the existing SuSiE-inf workflow.


R/univariate_pipeline.R
```diff
@@ -58,6 +61,7 @@ univariate_analysis_pipeline <- function(
     coverage = c(0.95, 0.7, 0.5),
     min_abs_corr = 0.8,
     finemapping_extra_opts = list(refine = TRUE),
+    fit_susie_inf = TRUE,
```
gaow: could you change this to `add_susie_inf` otherwise users might think we can only fit one model at a time not another.

PR #508, opened 2026-06-07
fix an error about missing sample size in sumstat file

xueweic:
The All methods parts is the fix of xqtl-protocol pipeline - enable `region_data$***`


gaow:
@danielnachun the unit test seem to pass. I will leave this for you to review and merge (after clarify with @xueweic on DM if needed; or approve it). I've left some comments.

NAMESPACE
```diff
@@ -261,6 +261,8 @@ exportClasses(MultivariateRegionalData)
 exportClasses(QCResult)
 exportClasses(RegionalData)
 exportClasses(TWASWeights)
+exportMethods("$")
```
gaow, line 266: @danielnachun do we need this?

R/AllMethods.R
```diff
@@ -252,6 +252,112 @@ setMethod("getGenotypeMatrix", "RegionalData", function(x) x@genotype_matrix)
 #' @export
 setMethod("getGenotypeMatrix", "MultivariateRegionalData", function(x) x@genotype_matrix)
 
+#' @export
+setMethod("$", "RegionalData", function(x, name) {
```
gaow, line 256: These seem useful but not clear to me of the necessity?

NAMESPACE
xueweic: @gaow @danielnachun this is related to xqtl protocol [mnm_regression](https://github.com/StatFunGen/xqtl-protocol/blob/main/code/mnm_analysis/mnm_methods/mnm_regression.ipynb). If we don't add this `$`, `fdat$residual_Y` for example will return an error since `S4vector`. We could either add `$` here, or change the xqtl protocol pipeline entirely based on S4 refactor.

R/AllMethods.R
xueweic: Same reason above. It would better if we can provide the transfermation, most people know more on `fdat$xxx` format than S4 factor.

danielnachun:
Superseded by https://github.com/StatFunGen/pecotmr/pull/509.

PR #515, opened 2026-06-11
Step 1.2: make maf optional (derived from af) in univariate and multivariate pipelines

Yining97:
`maf` is now optional in both `univariate_analysis_pipeline` and `multivariate_analysis_pipeline`, with `af` as the directional input. When `af` is given, the MAF used by `maf_cutoff` is computed from it (`min(af, 1-af)`); a passed-in `maf` is a fallback only and is never written to `top_loci`. Existing calls that pass `maf` behave the same.

This also closes a CI break from #513, which renamed `postprocess_finemapping_fits`'s `maf` argument to `af` but missed the multivariate caller (it still passed `maf` and failed `test_multivariate_pipeline.R`).

### DIFF
--- a/R/univariate_pipeline.R
+++ b/R/univariate_pipeline.R
@@ -8,8 +8,13 @@
 #' @param Y A vector of phenotype measurements.
 #' @param X_scalar A scalar or vector to rescale X to its original scale.
 #' @param Y_scalar A scalar to rescale Y to its original scale.
-#' @param maf A vector of minor allele frequencies for each variant in X. Used
-#'   only for \code{maf_cutoff} filtering; never exported.
+#' @param maf Optional vector of minor allele frequencies for each variant in X,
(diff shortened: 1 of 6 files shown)


R/univariate_pipeline.R
```diff
@@ -8,8 +8,13 @@
 #' @param Y A vector of phenotype measurements.
 #' @param X_scalar A scalar or vector to rescale X to its original scale.
 #' @param Y_scalar A scalar to rescale Y to its original scale.
-#' @param maf A vector of minor allele frequencies for each variant in X. Used
-#'   only for \code{maf_cutoff} filtering; never exported.
+#' @param maf Optional vector of minor allele frequencies for each variant in X,
+#'   used ONLY for \code{maf_cutoff} filtering and never exported. \code{af} is
+#'   the single source of truth: when \code{af} is supplied the filtering MAF is
```
gaow, line 13: @Yining97 please fix this as we discussed.

PR #518, opened 2026-06-13
RSS fine-mapping defaults: make L/lGreedy/minAbsCorr/rFinite/rMismatch optional

Yining97:
Step 3 of the RSS-parity plan. The RSS wrappers hardcoded fine-mapping parameters that `susie_rss` already defaults, so a pipeline run with none set did not equal a manual `susie_rss` run (the plan's pipeline-vs-manual success criterion).

## What changes
- **`susieRssPipeline()`**: `L`/`lGreedy`/`minAbsCorr` default to a NULL sentinel — omit-don't-substitute. `L`/`L_greedy`/`R_finite`/`R_mismatch` go into the `susie_rss` call only when set; `computeCsTable` omits `min_abs_corr` from `susie_get_cs` when unset, so susieR's own defaults apply. The `lGreedy <- min(lGreedy, L)` clamp runs only when both are set.
- **`rssAnalysisPipeline()`**: the `finemappingOpts` default drops `L`/`L_greedy`/`min_abs_corr`; absent keys forward as unset. `coverage`/`signal_cutoff` keep pipeline defaults.

## Testing
Modified package, susieR 0.16.4: 3093 pass, 0 fail, 22 skip. On chr21 AD_Bellenguez: explicit step-1 params reproduce the step-1 credible sets exactly (max abs PIP diff 0); all-unset matches a manual `susie_rss` run (max abs PIP diff 0); flagship explicit parity holds on shared keys incl. the 2nd CS at low `min_abs_corr`.

## Follow-up
`colocWrapper()` still sets its own `finemappingOpts` defaults (out of scope here). xqtl-protocol `rss_analysis.ipynb` should adopt the camelCase API and leave these params unset to inherit susieR defaults.

### DIFF
--- a/R/susie_wrapper.R
+++ b/R/susie_wrapper.R
@@ -257,6 +257,7 @@ postprocessFinemappingFits <- function(fits, dataX, dataY = NULL,
                                        region = NULL,
                                        priorEffTol = 1e-9,
                                        minAbsCorr = 0.8,
+                                       medianAbsCorr = NULL,
                                        csInput = NULL) {
   fits <- fits[!vapply(fits, is.null, logical(1))]
   if (length(fits) == 0) stop("At least one fine-mapping fit must be supplied.")
@@ -276,6 +277,7 @@ postprocessFinemappingFits <- function(fits, dataX, dataY = NULL,
       signalCutoff = signalCutoff, otherQuantities = otherQuantities,
       region = region,
       priorEffTol = priorEffTol, minAbsCorr = minAbsCorr,
+      medianAbsCorr = medianAbsCorr,
       csInput = csInput
     )
   })
@@ -343,6 +345,7 @@ postprocessFinemappingFit.susiF <- function(fit, method = "fsusie", csInput = NU
                                              region = NULL,
                                              priorEffTol = 1e-9,
                                              minAbsCorr = 0.8,
+                                             medianAbsCorr = NULL,
                                              csInput = c("X", "Xcorr", "fsusie")) {
   csInput <- match.arg(csInput)
   variantNames <- extractVariantNames(fit)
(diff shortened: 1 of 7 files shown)


R/susie_wrapper.R
```diff
@@ -940,19 +953,26 @@ adjustSusieWeights <- function(twasWeightsResults, keepVariants, runAlleleQc = T
 #' @importFrom dplyr arrange select
 #' @export
 susieRssPipeline <- function(sumstats, ldMat = NULL, xMat = NULL, n = NULL,
-                             L = 30, lGreedy = 5,
+                             L = NULL, lGreedy = NULL,
```
gaow: @Yining97 the point here is that L, L_greedy etc don't have to be part of the explicit pipeline parameter, if we allow for them to be passed in via `...`.

R/susie_wrapper.R
```diff
@@ -940,19 +953,26 @@ adjustSusieWeights <- function(twasWeightsResults, keepVariants, runAlleleQc = T
...
                              analysisMethod = c("susie_rss", "single_effect", "bayesian_conditional_regression"),
                              methods = NULL,
                              addSusieInf = TRUE,
                              coverage = 0.95,
                              secondaryCoverage = c(0.7, 0.5),
                              signalCutoff = 0.1,
-                             minAbsCorr = 0.8,
+                             minAbsCorr = NULL,
```
gaow: Including this, should be removed from the input. Your interface will be much simpler as a result.

R/susie_wrapper.R
```diff
@@ -1017,33 +1037,35 @@ susieRssPipeline <- function(sumstats, ldMat = NULL, xMat = NULL, n = NULL,
     shat <- sumstats$se
     names(bhat) <- names(shat) <- names(z)
     common <- c(list(bhat = bhat, shat = shat, var_y = varY, n = n,
-                     coverage = coverage, R_finite = rFinite,
-                     R_mismatch = rMismatch), dots)
+                     coverage = coverage),
+                if (!is.null(rFinite)) list(R_finite = rFinite),
```
gaow: This is extremely awkward and can totally be avoided if we use the `...` idea without specifying these from the parameter list

PR #539, opened 2026-06-30
Reconcile chr-prefix conventions in the LD-sketch lookup

Yining97:
`.ldFromSketch` matched requested variant ids against the panel with a raw string match, so genotype-LD fine-mapping (GWAS + QTL) errored with "variant id(s) not present in the LD sketch panel" whenever the entry ids and panel ids referred to the same variants but disagreed only on the `chr` prefix. pecotmr already ships `ensureChrMatch`/`normalizeVariantId` (variantId.R) but wired them in nowhere. This routes the lookup through `ensureChrMatch` before matching, preserving the caller's labels and the genuinely-absent error.

Tests: prefixed/non-prefixed both directions, same-convention regression, absent-still-errors; test_ld 469/0, test_fineMappingPipeline 385/0, test_sumstatsQc 799/0. Validated on the protocol_example chr22 MWE (common path unchanged; a stripped-prefix panel now resolves).

### DIFF
--- a/R/ld.R
+++ b/R/ld.R
@@ -548,7 +548,21 @@ loadLdFromGenotype <- function(genotypePath, region,
   }
   onMissing <- match.arg(onMissing)
   snpInfo <- getSnpInfo(ldSketch)
-  idx <- match(variantIds, as.character(snpInfo$SNP))
+  # Reconcile a pure chr-prefix mismatch between the requested ids and the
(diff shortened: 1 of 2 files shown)


danielnachun:
This is a good fix, I will merge it now.  I will also do an audit of the rest of the code to make sure this issue doesn't exist elsewhere.  chr-prefix mismatches are trivial to solve and should be handled automatically by the package.

PR #541, opened 2026-07-02
Add ser (single-effect) fine-mapping method for GWAS RSS

Yining97:
Adds a `ser` method token → `susieR::susie_ser` on the GwasSumStats/RSS path: an LD-free single-effect regression (takes z + n, no R, no L), distinct from `susie` with L=1. GWAS-only (a `gwasOnly` capability guard rejects it on QtlDataset/QtlSumStats with a clear message). Since `susie_ser` returns a susie-class object, it flows through the existing FineMappingEntry post-processing, with credible-set purity still computed against the LD sketch. Restores the legacy `univariate_rss` `single_effect` option to the current API.

Tests: ser dispatch → GwasFineMappingResult; rejected on QTL inputs; no L injected. test_fineMappingPipeline 394/0, plus wrappers/sumstatsQc/ld suites green. Validated with a real susie_ser fit on the protocol_example chr22 MWE.

### DIFF
--- a/R/fineMappingPipeline.R
+++ b/R/fineMappingPipeline.R
@@ -30,6 +30,11 @@
 #'           variant of the same.}
 #'     \item{\code{susieAsh}}{\code{unmappable_effects = "ash"}
 #'           variant of the same.}
+#'     \item{\code{ser}}{Single-effect regression via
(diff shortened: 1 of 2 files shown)


Yining97:
Closing — exposing ser isn't needed.

PR #571, opened 2026-08-11
Empty the LD sketch for zero-variant sumstats entries

Yining97:
### Problem
When a study has zero variants in the analysis window, its entry early-exits QC ("fewer than two variants after pre-harmonization QC") with `variantsIn = 0` — correct — but the retained `ldSketch` keeps the **full genome-wide panel** (13,205,438 rows, ~135 MB serialized) instead of being trimmed. Cause: `.subsetSketchToRange` and `.subsetSketchToIds` build their keep-set from the entries' variant positions/ids, so with zero variants they hit `if (length(...) == 0L) return(ldSketch)` and return the full sketch. Version-independent (0.6.9, 0.6.11). Results/audit are unaffected — only the serialized slot. Cost scales with the panel, not the region (a study missing 100 blocks writes ~13.5 GB of overhead).

### Fix
A zero-variant object references no LD, so its sketch should be empty. Added `.emptySketch()` (zeros `@snpInfo`, preserves all other slots) and made both trimmers empty on genuinely-zero variants. `.subsetSketchToIds` stays conservative (full sketch) only for the pathological "ranges present but `SNP` mcol absent" case, so a malformed-but-non-empty entry is never blanked. Populated studies are unchanged.

### Validation
- Zero-variant → 0-row sketch on both trimmers; emptied handle valid, `getNSamples` preserved, serializes/reloads. 200k-row synthetic sketch: 1.139 MB → 0.00052 MB (2199x smaller).
- Populated study byte-identical ([min,max] span / matched-id keep-set); conservative no-mcol case; NULL-safe.
- `test_sumstatsQc.R`: 328 tests, 0 failed (3 new tests, 16 expectations).

No API change; no change to results or the early-exit audit.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/R/sumstatsQc.R
+++ b/R/sumstatsQc.R
@@ -3132,13 +3132,33 @@ krigingOutlierQc <- function(zScore, R, n, variantIds = NULL,
   list(gr = .dfToEntryGranges(df), audit = entryAudit)
 }
 
+# Return `ldSketch` trimmed to zero variants (empty @snpInfo), preserving every
+# other slot (path / format / nSamples / sampleIds / chromPaths). Used when an
+# entry set genuinely carries no variants: the object references no LD, so the
+# correct retained panel is empty rather than the full genome-wide sketch. Edits
+# @snpInfo directly (not .subsetGenotypeHandle) because a zero-variant sketch is
+# never read for extraction, so the fileIdx read-safety guard does not apply --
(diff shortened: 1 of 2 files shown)


danielnachun:
This is now incorporated

